PR #1, opened 2015-03-19
Support of connecting via Unix-socket for MySQL checks

aleksey-hariton:
On my servers I have MySQL installed without networking support ("skip-networking" in my.cnf), so I need way to check MySQL via Unix-socket.
This commit allows to specify "--socket" option. Option supported in "dbi" and "mysqlcmd" sqlmode`s.

Example: 

```bash
./centreon_plugins.pl --plugin database::mysql::plugin --username='root' --host localhost --socket=/tmp/mysql1.sock --mode queries --sqlmode mysqlcmd
```


qgarnier:
Sorry for the time to applied your change. Thanks for it. I have commited on redmine. We have still the github and redmine. But in short time, we'll have only github.

PR #47, opened 2015-05-04
fix bug with unit for temperature and humidity (0.1 unit)

Shini31:
In MG-SNMP-UPS-MIB file, the humidity measurement is in unit 0.1 % (upsmgEnvironmentHumidity) and the temperature measurement is in unit 0.1 Cel (upsmgEnvironmentTemperature).


garnier-quentin:
Thanks.

PR #55, opened 2015-05-12
Ignore deleted poller

Gilum:
Add "WHERE deleted=0" in the request


garnier-quentin:
Thanks but could you put some quote ?
Like: deleted = '0' 

It could work if it's a enum.

Gilum:
Update quote.

garnier-quentin:
thanks!

PR #72, opened 2015-06-11
New mode SQL for postgres

s-duret:
Modification of plugin.pm and creation of sql.pm


garnier-quentin:
I refuse it. I'll do a dyn mode for that.

PR #73, opened 2015-06-15
Visimax

Sims24:
Add modes for visimax monitoring


garnier-quentin:
Need more info. Is it an appliance or a software on Windows ?

PR #76, opened 2015-06-20
Fix UNKNOWN error whan scenario not complet

Tpo76:
Fix UNKNOWN error whan scenario not complet


Tpo76:
Fix UNKNOWN error whan scenario not complet

garnier-quentin:
You have a timeout option for that.

PR #126, opened 2015-08-10
Issue92

Shini31:


garnier-quentin:
Need to use the new http lib please.

PR #134, opened 2015-08-14
Add Database Option

s-duret:


garnier-quentin:
sorry. Need to think globally. Because it can be used with more than one database.

PR #148, opened 2015-09-08
+ Fix qtree unit

Gilum:
snmp return KB unit


garnier-quentin:
Exact. Thanks :) But i have doned the change otherwise
777bb2ff3f683f1ec8eb2484718bd603babb38a2

PR #194, opened 2015-10-14
Rabbitmq - closed #157

### LINKED ISSUE #157: [plugin] RabbitMQ
Hello,

Can you create a plugin for RabbitMQ infrastructure ?

Regards.

Shini31:


garnier-quentin:
First could you put it in a subdirectory (like rabbitmq/restapi/). because i think we can also monitor it by JMX no ?

Shini31:
There are two ways to monitor RabbitMQ: either CLI or by API

garnier-quentin:
My bad. It's in Erlang. you get same information from the cli and rest api ?

Mode 'aliveness': help message seems wrong. Use system to system to overload status (could be useful).
Mode 'cluster': Don't use '$exit eq 'OK'' (there is a method for that). I don't understand the logic of the code (it goes OK and then critical ??!!).
Mode 'list': it works ???!! You can set multiple values like that ? Please use 'check_filter' (no more exclude old function).
Mode 'node': could you check all nodes at the same time ? Please use 'centreon::values' system.
Mode 'objects' and last one : Please use 'centreon::values' system.

For 'centreon::values' look: 
https://github.com/centreon/centreon-plugins/blob/master/centreon/common/aruba/snmp/mode/apusers.pm

Shini31:
Besides the error on the aliveness mode, everything else is 100% functional.
It's dramatic to have to redo all the code after doing testing by several people all for the beauty of the code especially when there is no documentation...

garnier-quentin:
It's not only for beauty. I need it to do refactoring after.

PR #211, opened 2015-11-19
+Add port monitoring for brocade

Sims24:
Adding advanced fiber port monitoring for brocade

+operState
+physicalState
+temperature

Regards


Sims24:
Huuum ... Looks like there is an huge fail on indentation, I close this one and will send an other request when cleaned !

PR #212, opened 2015-11-19
+Add port.pm 

Sims24:
[Add try2 after cleaning indentation]

+operstate
+physicalstate
+temperture

on brocade fiber channel ports

Thx


Sims24:
Will discuss together to find a way to add stats also, closing this one, will define all things tomorrow

PR #227, opened 2015-11-30
+ add Airspace (Cisco WLC) nb users by ssid

niclem:
Add new mode in plugin WLC (Aispace)
Get the number of users for each bssid configured on the Cisco WLC.


garnier-quentin:
I think i'll add it in ap-users mode. What do you think about it ?

niclem:
OK , nice !

PR #260, opened 2015-12-30
Add the possibility to monitor VPN with the plugin network::fortinet::fortigate::plugin.

sboulianne:
Hi Quentin,
It would be very great if you can add an option to monitor the VPN sessions in a FortiGate.

I found that but it wont look working.
https://exchange.nagios.org/directory/Plugins/Hardware/Network-Gear/Fortinet/Check-Fortigate-VPN-sessions/details

Since you are the BEST, you should be able to do something easily! ;)
I am available to test it when it will be done! :D

Thanks and merry christmas!


garnier-quentin:
Please open a ticket. Not a pull request :)

PR #270, opened 2016-01-07
Github: Add new mode to retrieve number of forks, stars and followers of a repository

bcournaud:


bcournaud:
I forgot to add the values in the output (it's only in the perfdata).
So I will close this one and make a new pull request

PR #276, opened 2016-01-12
Add perl-Date-Manip in Centos/Rhel 6 doc

fixlair:
Date/Manip.pm is needed for --plugin=apps::protocols::x509::plugin --mode=validity for example


garnier-quentin:
Hi fixlair,

For which plugins do you need that ?

fixlair:
Hi,
I need it for X509 plugin with mode validity 
Regards,

garnier-quentin:
Thanks for the feedback. DateManip is a little overkill. 
I think i won't put it in doc (it's only for the main dependencies. Like i don't put oracle dbd for example). 
I think we change it to use a "easiest" cpan module like "Date::Parse".

PR #278, opened 2016-01-12
Fix #277

### LINKED ISSUE #277: [Mode] Add support alternative subject names in X509's validity mode
Hello,

Can you add the support of alternative subject names in X509's validity mode

Regards.

Shini31:


garnier-quentin:
Thanks. I think you should display all subject names matched your regexp (and not the regexp). What do you think ?
Right now, you display the regexp. In the debug long output, you could display subject names which have not matched.

Shini31:
I just add it

PR #290, opened 2016-01-18
Add the possibility to monitor the status of the Microsoft DFS.

sboulianne:
Hi,

It will be great if we can monitor the status of the DFS on Microsoft servers and check if all is working perfectly (replication, services, etc).
I found this script but it didn't look complete.
https://exchange.nagios.org/directory/Addons/Monitoring-Agents/Microsoft-DFS-Monitor/details

Thanks you very much in advance.


garnier-quentin:
Sorry but it shouldn't be opened as a pull request.

PR #298, opened 2016-01-22
Added new apcupsd format

florian-asche:


garnier-quentin:
Why do you put '|C' two times in your regexp ?

florian-asche:
Sorry, my mistake, fixed it.

Issue #354, opened 2016-03-09
hardware::pdu::schleifenbauer::gtw::snmp::plugin new plugin to check schleifenbauer PDU's


cemir:
Hi,

here is a new plugin for Schleifenbauer hardware,

currently it's able to talk with the gateway through SNMP to check a PDU for status, humidity and temperature

power metrics will be added later.

Regards

garnier-quentin:
Here ? I don't see it :)
Could you give me the MIBs Files and a complete snmpwalk also:
snmpwalk -Cc -v2c -c public -ObentU localhost 1.3.6 > myagent.snmpwalk

Thanks!

cemir:
Actually I wrote it but i can't manage how to pull request
I followed what's on this page https://documentation.centreon.com/docs/centreon-plugins/en/latest/developer/guide.html but I think it's outdated and was for the forge

garnier-quentin:
You can attach your code if you cannot pull request. But i need mibs and snmpwalk ;)

cemir:
Ok, I think i was able to create a pull request,

do you still need the snmpwalk ?

Regards,

garnier-quentin:
The MIBs files and the walk yes!

cemir:
command used was snmpwalk -Cc -v1 -c XXX 192.168.2.36 1.3.6 > myagent.snmpwalk

because snmp server supports only v1 and crash with -ObentU

[myagent.snmpwalk.txt](https://github.com/centreon/centreon-plugins/files/165321/myagent.snmpwalk.txt)
[SPGWMIB124.zip](https://github.com/centreon/centreon-plugins/files/165325/SPGWMIB124.zip)

garnier-quentin:
Thanks

cemir:
Hi, any news ? I created a pull request for this but no reaction on it ... Do I have to change/improve something ?

garnier-quentin:
Sorry for the time. First thanks. But i can't accept the PR. For example the temperature:
* you check only the 'sensor2Value'. But it could be 'sensor1Value' or 'sensor3Value',... Moreover you use an argument for the device ID and you don't check if there are many 'channelIndex8'. 

[...]

PR #355, opened 2016-03-09
Added probes for Schleifenbauer PDU's ref #354

cemir:


garnier-quentin:
No response about it: #354
I let the issue open

PR #392, opened 2016-05-02
Update uptime.pm

benoitb21:


garnier-quentin:
Thanks for the pull request. I have used a function to display it. Is the result ok ?

benoitb21:
Yes but I thinks that is better if we can choose to limit the "days display" because after it's an average operation. Y (31556926s) => 365.24 days, M (2629743s) => 30.44 days. 
It's not a reality, because each month and year are differents. But, it's my opinion.

garnier-quentin:
I have done the change. Is it better ?

benoitb21:
Perfect :)

PR #413, opened 2016-05-25
display "total" perfdata counter

jazzl0ver:
this makes the disk usage graph easier to read and understand


garnier-quentin:
Thanks for the pull request.
But i think i need to enhance the option --explode-perfdata-max with instance. To get something more generic with instance.

jazzl0ver:
Thanks for the hint! I really missed this option.

PR #421, opened 2016-06-03
2.7.x

Oly75:
Subject : Virt-Vmware2

Hi Quentin,

I would like to know if it's possible to get the number of Vm running on one Vcenter. I know can check by datastore or by esx. But I would to know how many Vm is runing on my Vcenter. To have this kind of output : 	
runtime OK - 227/236 VMs up, 26/28 Hosts up, Farman - Ulis overall status=gray, no config issues.
Like I have with the check_vmware_api plugin.

Regards


garnier-quentin:
Indeed it's by ESX. Need also a total. But it's not the good repository.

PR #443, opened 2016-07-14
fix traffic for RHEL7

phoppe93:
the network check is not working with rhel7 systems.

RHEL7 ifconfig:
RX packets 130978614  bytes 93552141514 (87.1 GiB)
RX errors 0  dropped 86  overruns 0  frame 0
TX packets 123621691  bytes 109981907800 (102.4 GiB)
TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

RHEL6 ifconfig:
RX packets:15109919 errors:0 dropped:0 overruns:0 frame:0
TX packets:13222273 errors:0 dropped:0 overruns:0 carrier:0
collisions:0 txqueuelen:1000
RX bytes:8451113127 (7.8 GiB)  TX bytes:28249980454 (26.3 GiB)

The regex is wrong because the output of ifconfig is changed.

### DIFF
--- a/os/linux/local/mode/traffic.pm
+++ b/os/linux/local/mode/traffic.pm
@@ -128,8 +128,25 @@ sub manage_selection {
         next if (defined($self->{option_results}->{name}) && !defined($self->{option_results}->{use_regexp}) && !defined($self->{option_results}->{use_regexpi})
             && $interface_name ne $self->{option_results}->{name});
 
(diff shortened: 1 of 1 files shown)


garnier-quentin:
Thanks for the PR. Could you test it ?

phoppe93:
yes it works!

garnier-quentin:
Thanks for the feedback!

PR #456, opened 2016-08-02
Round remaining time to minutes

jframeau:
This is to avoid to display remaining time like: Remaining time : 460.366666666667 minutes

NB: i'm not a perl guru, so using sprintf to round a float is perhaps not the right way to do.

jfr


garnier-quentin:
Thanks for the PR. Could you add some parenthesis for sprintf ?

PR #460, opened 2016-08-05
Openstack

### LINKED ISSUE #156: [plugin] Openstack
Hello,

Can you create a plugin for Openstack infrastructure ?

Regards.

Shini31:
closed #156

### DIFF
--- a/cloud/openstack/restapi/custom/osapi.pm
+++ b/cloud/openstack/restapi/custom/osapi.pm
@@ -191,7 +191,7 @@ OpenStack Rest API custom mode
 
 =item B<--data>
 
-Set file with JSON request
+Set file with JSON request (see http://docs.openstack.org/developer/keystone/api_curl_examples.html for more explanations)
 
 =item B<--hostname>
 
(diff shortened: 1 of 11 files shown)


garnier-quentin:
I cannot accept it. First, it could be good to merge in one mode 'components' for the following modes: 'hypervisor', 'instance', 'network', 'port' and 'volume'.

But i need information about API performance (time for API calls). How i see the API:
* list tenants. It seems, we cannot use wildcard for we'll add an option to filter one tenant
* '/v2.1/​{tenant_id}​/os-hypervisors': status 
* '/v2.1/​{tenant_id}​/servers/detail': status
* '/v2/​{tenant_id}​/volumes/detail': status
* '/v2.0/networks': status (tenant id is in the response)
* '/v2.0/ports': status (tenant id is in the response)

With so many tenants, we should use following syntax to filter attributes: GET 'fields=id&fields=name' 
It could enhance performance.

If someone could give me answer about performance, it would be much appreciate.

PR #462, opened 2016-08-10
Service status was everytime the last status in %bitmap_status

s-duret:


garnier-quentin:
i'll look the mib file but it's a bitmask. You cannot use equals. But i think i can enhance because i get only one status.  Maybe we can have X status.

PR #498, opened 2016-09-21
Adds new expected string options

tanguyvda:
Allows you to check for a warning string or a critical string on the webpage


garnier-quentin:
I see the idea. But i prefer another way

PR #532, opened 2016-11-04
Update sonus plugin package namespace

micsco:
Sonus plugin had an incorrect package namespace which was causing it to load incorrectly.

### DIFF
--- a/network/sonus/sbc/plugin.pm
+++ b/network/sonus/sbc/plugin.pm
@@ -18,7 +18,7 @@
 # limitations under the License.
 #
 
-package network::sonus::sbc::snmp::plugin;
+package network::sonus::sbc::plugin;
 
 use strict;
 use warnings;


garnier-quentin:
It should be better. Could you test it ?

garnier-quentin:
It should be better. Could you test it ?

garnier-quentin:
It should be better. Could you test it ?

garnier-quentin:
It should be better. Could you test it ?

garnier-quentin:
It should be better. Could you test it ?

micsco:
Indeed, sorry didn't truly appreciate the directory structure etc you were going for. Just dipping my toes - thank you!

Sims24:
Hi @micsco , 

We're still looking for hardware sonus equipment to add temperature and hardware checks. If you have such device, we would be glad to enhance this plugin with new modes ;) 

Thanks!

PR #536, opened 2016-11-17
Add support of Sophos XG UTM

rk4an:


rk4an:
Just see that a Cyberoam plugin was added few day ago. Note that Sophos XG is based on Cyberoam and will evoluting with maybe a new MIB from Sophos

https://github.com/centreon/centreon-plugins/releases/tag/20161014

garnier-quentin:
You cannot use the cyberoam plugin instead ? If you are some modes missing, we can add it in cyberoam.
And then when a new mib coming, we'll add a plugin for Sophos XG

rk4an:
We can do that but some oid of the Cyberoam mib doesn't work on Sophos (hit for example). I think we can already split it (cyberoam plugin for old device and sophos for new one)

garnier-quentin:
Ok i see.

rk4an:
I close this pull request and reopen a new one (problem with commit)

PR #550, opened 2016-11-25
Sophos XG plugin

rk4an:


garnier-quentin:
I have added live users in 'requests' mode and i have created the mode 'services'. It's a cyberoam appliance. So i don't think useful to create a XG plugin. What do you think about it ?

rk4an:
Hi, ok I think we can do that for the moment, until Sophos release a specific MIB for the XG. Thanks

PR #569, opened 2016-12-22
To remove the line breaks in the description.

jben94:
Example:

```

          '.1.3.6.1.2.1.47.1.1.1.1.2.4954' => 'FixedModule-1 Port-5',
          '.1.3.6.1.2.1.47.1.1.1.1.2.21598' => 'PowerSupply-1 Sensor-1
',
          '.1.3.6.1.2.1.47.1.1.1.1.2.4953' => 'FixedModule-1 Port-4',
          '.1.3.6.1.2.1.47.1.1.1.1.2.149' => 'Nexus5020 Chassis',
```

capture

### DIFF
--- a/centreon/common/cisco/standard/snmp/mode/environment.pm
+++ b/centreon/common/cisco/standard/snmp/mode/environment.pm
@@ -123,8 +123,8 @@ sub snmp_execute {
     
     push @{$self->{request}}, { oid => $oid_entPhysicalDescr }, { oid => $oid_ciscoEnvMonPresent };
     $self->{results} = $self->{snmp}->get_multiple_table(oids => $self->{request});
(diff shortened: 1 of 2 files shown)


garnier-quentin:
Could you use trim function from centreon::plugins::misc ?

jben94:
No problem, but I'll have to change it too. Because it does not take my case into account.

garnier-quentin:
The use is missing at the beginning of the file ?

jben94:
Sorry,

Not being at the same version as you, I can not test.
I use "centreon::plugins::misc", like you with

```
centeron::plugins::misc::mymodule_load (output => $ self -> {output}, module => $ mod_name,
                                                    Error_msg => "Can not load module '$ mod_name'.");
```

In an old version.

PR #592, opened 2017-02-06
Add raisecom plugin

cgagnaire:


maksimatveev:
@cgagnaire Thank you! I was waiting for that will be able to finish! =)

garnier-quentin:
You need to do some fix:
* hardware.pm: you have filter on psu and fan. But it seems there is nothing for that hardware in the mib. Could you add voltage also ?
* temperature.pm: could you use threshold from the mib like 'raisecomTemperatureThresholdLow' ?
* cpu.pm: i don't like how you fix the values from OIDs for cpu. Nothing say that it will be fixed between equipments. Please use 'raisecomCPUUtilizationPeriod' and value 'onesec', 'fivesec',... to be something more consistent
* memory.pm: seems ok for me.

Thanks for the work

garnier-quentin:
There is also some fanmonitor information (status, speed and temperature). Please add it in hardware mode. It's in mib 'RAISECOM-FANMONITOR-MIB.txt'.
You also have powermonitor information. In mib 'RAISECOM-POWERMONITOR-MIB'.

There is also have some hardware for PON series. Should check also that. it's for iscom5800 for example.

cgagnaire:
I added voltage and fan components for hardware mode.
I added the default thresholds for temperature component.
I can't find anything interesting in 'RAISECOM-POWERMONITOR-MIB'.
For CPU, I'm not sure to fully understand what you're saying : the indexes will always be the same as describe in the MIB (INTEGER {oneSec(1), fiveSec(2), oneMin(3), tenMin(4), twoHour(5)}), but what you're saying is that it may happen that one of those five values will be undefined ?

garnier-quentin:
The value 'oneSec(1), fiveSec(2), oneMin(3), tenMin(4), twoHour(5)' in parenthesis is not the index. It's the value for the type OID.

cgagnaire:
Well, it represents the periods so 'raisecomCPUUtilization.4' will always be the CPU utilization for the last ten minutes right ?

maksimatveev:
@cgagnaire my version of the plugin I just use these values of OIDs.
https://github.com/maksimatveev/centreon-plugins/tree/master/network/raisecom/snmp

PR #634, opened 2017-04-06
add plugin HighSMS

benrobert:
new plugin highsms


garnier-quentin:
Seems fine. But:
- could you add proxy options ? Should be useful i think. 
- could you remove Data::Dumper
- protect the text in xml with ```<![CDATA[]]>``` syntax

Thanks!

PR #699, opened 2017-07-07
Take into account non HTTP samples

sebastien-prudhomme:
XML output is slightly different when using non HTTP scenarios with JMeter (for instance for a database test plan).

The attributes of the XML nodes are still the sames.


garnier-quentin:
thanks

PR #706, opened 2017-07-12
Fix oid mapping

yodaaut:
this module should go for raidDiskState instead of raidDiskSyncState and i added some possible values


garnier-quentin:
Thanks for the PR. But could you let 'failed' on following line ?
3 => 'disc_failed',

yodaaut:
i can, but it's officially called "disc_failed" in the MIB

garnier-quentin:
On my mib it's 'failed' :) They have change it maybe. If you put disc_failed, change in hardware.pm also.

PR #710, opened 2017-07-18
sms with foxbox modem

lkappel:
usage : 
perl  ./centreon_plugins.pl --plugin=notification::foxbox::plugin --mode='alert' --from='centreon' --hostname '10.40.1.124' --username 'admin' --password 'centreon' --phonenumber '33612345678' --testo 'Service Alert: Central/proc-ntpd \nService State: CRITICAL \nDate/Time: Thu Jul 6 18:58:48 CEST 2017'

### DIFF
--- a/notification/foxbox/mode/alert.pm
+++ b/notification/foxbox/mode/alert.pm
@@ -26,10 +26,6 @@ use strict;
 use warnings;
 use centreon::plugins::http;
 
-use constant OK        => 0;
-use constant WARNING   => 1;
-use constant CRITICAL  => 2;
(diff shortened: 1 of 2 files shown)


garnier-quentin:
My remarks:
* should --texto or --sms no ? 
* could you use outpu methods to set text in output ? 
* could you use an option despite a constant for the url path ?

garnier-quentin:
My bad. it's not ok. You need to use the library http. Otherwise, you also don't manage output well. You can use directly 'UNKNOWN' in output_add. The exit is "smart" about the exit code according what you put in output_add.

PR #730, opened 2017-08-02
Fixed --help options no working with --dyn-mode

Orabig:
There was no help available while using dyn-mode.
This patch fixes this


garnier-quentin:
Thanks. I add it for other lib.

PR #753, opened 2017-09-06
Postgres add databasesize

Tpo76:
add database size mode


cgagnaire:
Hi @Tpo76,
As you can have multiple databases, you should look for all and use filters like for https://github.com/centreon/centreon-plugins/blob/master/database/mysql/mode/databasessize.pm and https://github.com/centreon/centreon-plugins/blob/master/database/mssql/mode/databasessize.pm.
Try to also use the centreon::plugins::templates::counter librairy.
Thank you :)

Sims24:
Following up on this https://github.com/centreon/centreon-plugins/pull/870 

Close this one

cgagnaire:
Just saw that https://github.com/centreon/centreon-plugins/blob/master/database/mysql/mode/databasessize.pm is not a good example. Use the other one !

PR #808, opened 2017-11-27
Show used and free values in the same probe

jben94:
Hi,

I open this PR adding the possibility to have the free space values in the same probe.
What is interesting when we want to make reports ....


garnier-quentin:
Thanks but i can't accept it. The threshold won't be good. Moreover, it must be more global.

jben94:
Hi,

Ok, I will review my changes. 
But can you add this option, to have both values in one check?

garnier-quentin:
It's too specific.

jben94:
So you have to check twice to have the used and free values in order to use usage rate (BI) reports?

Sims24:
That's a main change. We know that it's important for a lot of our user, but we do not want to implement it in plugins. There's a better place ;) 

Thanks

PR #841, opened 2017-12-28
improve redis restapi

cgagnaire:
- Add more counters,
- Add interval option to select the interval from which to retreive the statistics,
- Remove statefile as it is not used.


garnier-quentin:
The interval is for all modes ? If so, could you add it in custom api ?

PR #843, opened 2017-12-29
improve redis restapi add list modes

cgagnaire:


garnier-quentin:
thanks. Could you rename it like that ?
```
'list-nodes'         => 'apps::redis::restapi::mode::listnodes',
```

cgagnaire:
I also renamed the other modes

PR #844, opened 2017-12-29
improve apps vmware nethost

cgagnaire:


cgagnaire:
Goes along with https://github.com/centreon/centreon-vmware/pull/53

PR #852, opened 2018-01-10
Add list-storage mode for windows NSClient

Tpo76:


garnier-quentin:
It woul be better to create a dll in nsclient i think ;)

Tpo76:
Does it make sense to create dll in NSClient to have diso-show option ?
If yes, I can check that

garnier-quentin:
An dll to add centreon commands (not disco show option).

Sims24:
Looks good to me. I merge it. Please think about centreon_plugins.exe building and also discovery rules.

PR #854, opened 2018-01-11
Ntlmv2

pkriko:
Add support of NTLMv2 authentication.

You will need to install perl-NTLM-1.09-5.el7.noarch from EPEL repository to use the 1.09 version of Authen::NTLM to support V2 authentication.


garnier-quentin:
You can look "files changed", the indent is not good. Also, please use following function to load the module:
```
centreon::plugins::misc::mymodule_load(output => $self->{output}, module => 'Memcached::libmemcached',
error_msg => "Cannot load module 'Memcached::libmemcached'.");
```

PR #860, opened 2018-01-17
improve authentification handling

cgagnaire:
Subclassed _get_basic_credentials_ so it gives back the credentials provided in command line, and doesn't store them in the basic_authentication hash (http://search.cpan.org/~ether/libwww-perl-6.31/lib/LWP/UserAgent.pm#get_basic_credentials). 
This way, requests can be done on web servers asking for Basic, NTLM, NTLMv2 or Digest authentification.

Authen::NTLM > 1.04 is still needed for NTLMv2 authentification (cf @pkriko commits https://github.com/centreon/centreon-plugins/commit/d33a1c217eabc63158cab5e38a3cd56526940dac).

Option _--ntlm_ is now useless but remains for compatibility.

In case of _401_ error, the output will show you the expected authentification type.

### DIFF
--- a/apps/protocols/http/mode/expectedcontent.pm
+++ b/apps/protocols/http/mode/expectedcontent.pm
@@ -198,10 +198,6 @@ Set path to get Webpage (Default: '/')
 
 Specify this option if you access webpage over basic authentication
 
-=item B<--ntlm>
-
-Specify this option if you access webpage over ntlm authentication (Use with --credentials option)
-
 =item B<--ntlmv2>
 
 Specify this option if you access webpage over ntlmv2 authentication (Use with --credentials and --port options)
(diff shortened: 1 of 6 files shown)


garnier-quentin:
no our please ;)

cgagnaire:
@garnier-quentin,
I think it is better now.

garnier-quentin:
It's better yes. But when i look the doc:
```
# This checks wantarray and can either return an array:
    my ($user, $pass) = $ua->get_basic_credentials( $realm, $uri, $isproxy );
    # or a string that looks like "user:pass"
    my $creds = $ua->get_basic_credentials($realm, $uri, $isproxy);
```

So we should manage return types (array or string) with wantarray.

cgagnaire:
Done

garnier-quentin:
If the option remains for compatibility, you can still remove it from help pod.

PR #862, opened 2018-01-18
improve mssql blockedprocesses waittime

cgagnaire:
Improve MSSQL blockedprocesses mode to add the notion of time since when the blocked process is waiting. The blocking process is also shown.


cgagnaire:
Will be refactored to use centreon::plugins::templates::counter class...

Sims24:
Think that @omercier is handling this now. He will create the PR then. 

Thanks

Issue #210, opened 2015-11-18
network::cisco::standard::snmp::plugin packet-errors


kadrika:
Hi,

Can you please add packet-error mode to network::cisco::standard::snmp::plugin?

garnier-quentin:
It is Kadrika. The new mode 'interfaces' can do it with option: --add-errors option.

kadrika:
Ok, now I found it :) Thanks!
One more question is it possible to add exclusion option?

I have routers where I need to check errors in all ports, but the port where ISP is connected has errors and we cant do anything about it. So I would like to exclude that one port from the packet-error check. Is it doable?

garnier-quentin:
You can select the interfaces you need to check.

kadrika:
So no way to have one service to check all interfaces packet-errors and just exclude one interface?

So right now I need to add packet-error service to all interfaces, for example if I have 48 port switch I need to add each one separately? I instead 1 service I will have 48 services in monitoring?

garnier-quentin:
Yes you can exclude one interface by name. With negative regexp like : '^(?!(toto)$)'

kadrika:
Thank you! Thank you !! :)

kadrika:
But no way to exclude more than one? For example if I need to exclude fa0/13 and fa0/14 ?

Sims24:
Hi,

'^(?!(toto|tata)$)' will exclude toto and tata

kadrika:
And for the critical threshold I right now use --critical-in-error=10 --critical-out-error=10 --critical-in-discard=10 --critical-out-discard=10 . Is there a shorter way to use the critical threshold, if I want all of them to be the same, in this case 10.

garnier-quentin:
Nop. Need to do it for all.

kadrika:
Ok, Thanks!

PR #867, opened 2018-01-24
Snmp interfaces enhancement

proxyconcept:
I know we can use negative regexp to exclude interfaces (see for example https://github.com/centreon/centreon-plugins/issues/210 ), but it's not very practical...

Negative regexp are not intuitive, complex to write and to read, difficult to integrate into a centreon command. So I think that an additional option dedicated to exclusion is a  good idea !

### DIFF
--- a/snmp_standard/mode/interfaces.pm
+++ b/snmp_standard/mode/interfaces.pm
@@ -658,6 +658,7 @@ sub new {
                                 "global-admin-down-rule:s"  => { name => 'global_admin_down_rule', default => $self->default_global_admin_down_rule() },
                                 "global-oper-down-rule:s"   => { name => 'global_oper_down_rule', default => $self->default_global_oper_down_rule() },
                                 "interface:s"             => { name => 'interface' },
+                                "exclude:s"               => { name => 'exclude' },
                                 "units-traffic:s"         => { name => 'units_traffic', default => '%' },
                                 "units-errors:s"          => { name => 'units_errors', default => '%' },
                                 "speed:s"                 => { name => 'speed' },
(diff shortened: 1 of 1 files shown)


garnier-quentin:
Why it's difficult to integrate in centreon ? You use a service macro in your command and it's ok.

proxyconcept:
It's more difficult because the operators who add the services checks must master complex regexp syntax. Also it's more difficult to create generic centreon's commands and service templates with optionnal parameters.

For example you can use : --name --interface='^Gi0/(?!(1|5|8)$)'
But it could be easier to use : --name --interface='^Gi0'  --exclude='/(1|5|8)$'

With my enhancements operators can define a filter "all but a few exceptions" without any regexp syntax. For example : --name --exclude="Gi0/1,Gi0/5,Gi0/8" (works also without --name and a list of indexes)

proxyconcept:
To be complete, I just added another commit to consider empty values (for options --interface and --exclude) as undefined. With this we can create centreon's command with optional arguments like this : --interface='$ARG1$' --exclude='$ARG2$'
Then you can decide later in your service if you want all interfaces or not.

Note that currently, documentation says : "empty means 'check all interface'". But this is true only if you consider "empty" as "missing option". If you really run the command with an empty value (--interface=''), all interface are selected only with the --name option, else you get a "No entry found".

Sims24:
Hi, 

Thanks for your contribution.  

I would prefer to manage a flag option like "--interface="regexpexpr" --regexp --exclusion"K 

That might look overkill but we will avoid unwanted behaviour and trouble during upgrade, also easier to manage this through Centreon configuration templating system. 

Can you change PR according to this ?

PR #870, opened 2018-01-29
Add database-size mode

Tpo76:


Sims24:
Hi @Tpo76 

Thanks, here is the things you may change before we merge this : 

- Use counter class, look a mysql database/table size mode, this guy did it :p 
- Use fetchrow_hashref and will loop, more readable code when using $row structure content
- Indent regarding guidelines 

See you soon !

Issue #780, opened 2017-10-16
Cisco spanning tree add options to select interface and status


lpinsivy:
For Cisco spanning tree, add the possibility to select interface with spanning tree and an option/filter to overwrite status.

For example, we can disable a network interface on a switch to prevent loop. So it is normal that this network interface is down on a switch.

nailyk-fr:
I agree! 
Gonna **try** to work on a patch. 
Will you be able to test @lpinsivy ?

Sims24:
Hi there, 

We skip disable port by design, and also we consider as OK if spt state is disabled.

What would you want to change about the following rules : 

```
    1 => ['disabled', 'OK'], 
    2 => ['blocking', 'CRITICAL'], 
    3 => ['listening', 'OK'], 
    4 => ['learning', 'OK'],
    5 => ['forwarding', 'OK'],
    6 => ['broken', 'CRITICAL'],
    
    10 => ['not defined', 'UNKNOWN'], # mine status
```

@nailyk-fr you would wish to override some of the mapping spt state -> status described above ? 

Regards

nailyk-fr:
> @nailyk-fr you would wish to override some of the mapping spt state -> status described above ?

thanks for fastiness. 
Most likely something to overwrite this: 
` if (${$states{$stp_state}}[1] ne 'OK') {`
with custom values (like with `--stp-overwrite-status 21:disabled --stp-overwrite-status 22:blocked`) to be "in-line" with network STP topology. 
Example, if states change from `blocked` to `forwarding` STP have changed somewhere and should be displayed. 

P.S. am trying to change the plug-in, and guess the hardest part will be to retrieve the custom values.

Sims24:
Then you want to catch state transitions ? Or gain flexibility about status and ports (I assume 21: and 22: corresponds to a port number right ?) ? 

This plugin should be refactor with the new counter template class anyway. 

You may want to get some example with the bgppeerstate.pm which is probably close from what you try to achieve.

nailyk-fr:
Exactly. Catch status transition. 
Thanks will look at the BGP plugin for code samples. 

Am not able to help you on the plugin rewrite. Is it ok if I add this functionality on the old plugin anyway ?

Sims24:
We would prefer that you propose a PR according refactoring, but as we miss some time if your change keep backward compatibility with existing check (adding instead of breaking/changing, especially options) we may merge it of course.

nailyk-fr:
[...]

PR #912, opened 2018-03-20
SNMP: STP: Allow custom topologies

nailyk-fr:
Those PS will allow custom topologies, as mentioned in #780 

This is not perfect as it: 

* Output https://github.com/centreon/centreon-plugins/compare/master...nailyk-fr:master#diff-be436d687f581cbfaf7fa9f480ae2287R222 is not using the new way as it seems the new way cannot handle critical states, 
* Guess https://github.com/centreon/centreon-plugins/compare/master...nailyk-fr:master#diff-be436d687f581cbfaf7fa9f480ae2287R221 probably can cause false negative status, but cannot find in what condition, 
* As @ldidry suggest to me https://github.com/centreon/centreon-plugins/compare/master...nailyk-fr:master#diff-be436d687f581cbfaf7fa9f480ae2287R193 `(scalar $#stp_overwrite_status)+1` is now replaced with `(scalar @stp_overwrite_status)` as it is more readable, accurate, and do not require the `+1` which can lead in miss-reading the array. (same here https://github.com/centreon/centreon-plugins/compare/master...nailyk-fr:master#diff-be436d687f581cbfaf7fa9f480ae2287R162 but does not really matters)

Thanks :)

CC: @Sims24

### DIFF
--- a/snmp_standard/mode/spanningtree.pm
+++ b/snmp_standard/mode/spanningtree.pm
@@ -102,6 +102,7 @@ sub new {
                                 "filter-port:s"         => { name => 'filter_port' },
                                 "warning-status:s"      => { name => 'warning_status', default => '' },
                                 "critical-status:s"     => { name => 'critical_status', default => '%{state} =~ /blocking|broken/' },
+                                "stp-overwrite-status:s"=> { name => 'stp_overwrite_status' },
                                 });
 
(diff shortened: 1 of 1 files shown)


Sims24:
Closing this one. Thanks for your contribution !

PR #929, opened 2018-03-27
Re-calculating Virtual Memory for Windows machines

jmpalomares:
For some reason, the virtual memory for windows hosts are not correctly defined in windows machines with SNMP. When you do a request of OID .1.3.6.1.2.1.25.2.3.1.5 instead of gives you the INTEGER for the Virtual memory only, it gives you the sum of all the memory in the System (physycal + virtual)

I've edited the Swap module for windows OS in order to re-calculate the Virtual Memory correctly substracting the physical memory from the virtual memory (.1.3.6.1.2.1.25.2.3.1.5 and .1.3.6.1.2.1.25.2.3.1.6) in order to get the real Virtual Memory.

### DIFF
--- a/os/windows/snmp/mode/swap.pm
+++ b/os/windows/snmp/mode/swap.pm
@@ -29,16 +29,16 @@ sub new {
     my ($class, %options) = @_;
     my $self = $class->SUPER::new(package => __PACKAGE__, %options);
     bless $self, $class;
-    
+
     $self->{version} = '1.0';
     $options{options}->add_options(arguments =>
-                                { 
+                                {
                                   "warning:s"               => { name => 'warning' },
                                   "critical:s"              => { name => 'critical' },
                                 });
 
     $self->{swap_memory_id} = undef;
-    
+
     return $self;
 }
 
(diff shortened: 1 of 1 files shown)


Sims24:
Hi @jmpalomares 

Thanks A LOT for helping us improving centreon_plugins. 

We're aware of this problem for a while since SNMP on windows is shitty ... for a while :)

Our decision is to strongly recommend to use our client plugins  (nrpe or restapi) along the NSClient++ agent  to monitor MS environment. Page file usage will be more accurate for such monitoring. 

We may address the problem as you did through this PR, but we would prefer that Windows fix its own issue and not hidding such issue with code.  

Anyway, I could understand that it's a problem to you to use an agent instead of SNMP. If so, it would be really cool to adapt yout PR by adding an option letting users choose to bypass or not this windows bug. (Also not changing running behaviour of the plugin that could trigger false-positives somewhere). 

IMOO, --wash-ms-hands would be a really hype name for such option =)

cgagnaire:
Thanks @jmpalomares, I will add the option.

PR #932, opened 2018-03-27
Fix bad character f5 vs and pool

cgagnaire:
Fix bad first character in long and short outputs of virtual servers and pools name of F5 BigIP plugin.

### DIFF
--- a/network/f5/bigip/snmp/mode/poolstatus.pm
+++ b/network/f5/bigip/snmp/mode/poolstatus.pm
@@ -181,8 +181,7 @@ sub manage_selection {
     my $snmp_result = $options{snmp}->get_multiple_table(oids => [
                                                             { oid => $mapping->{new}->{AvailState}->{oid} },
                                                             { oid => $mapping->{old}->{AvailState}->{oid} },
-                                                         ],
-                                                         , nothing_quit => 1);
+                                                         ],  nothing_quit => 1);
     
     my ($branch_name, $map) = ($mapping->{new}->{AvailState}->{oid}, 'new');
     if (!defined($snmp_result->{$mapping->{new}->{AvailState}->{oid}}) || scalar(keys %{$snmp_result->{$mapping->{new}->{AvailState}->{oid}}}) == 0)  {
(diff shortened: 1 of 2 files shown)


network/f5/bigip/snmp/mode/poolstatus.pm
```diff
@@ -197,8 +196,9 @@ sub manage_selection {
         my $result = $options{snmp}->map_instance(mapping => $mapping->{$map}, results => $snmp_result->{$branch_name}, instance => $instance);
         $result->{Name} = '';
         foreach (split /\./, $instance) {
-            $result->{Name} .= chr  if ($_ >= 32 && $_ <= 126);
+            $result->{Name} .= chr if ($_ >= 32 && $_ <= 126);
         }
+        $result->{Name} =~ s/^.//;
```
Sims24: Looks really odd to fix it this way.   Do we have on the root cause of this first random char ? Is this chr provoking it

network/f5/bigip/snmp/mode/virtualserverstatus.pm
```diff
@@ -187,8 +186,10 @@ sub manage_selection {
         
         $result->{Name} = '';
         foreach (split /\./, $instance) {
-            $result->{Name} .= chr  if ($_ >= 32 && $_ <= 126);
+            $result->{Name} .= chr if ($_ >= 32 && $_ <= 126);
         }
+        $result->{Name} =~ s/^.//;
```
Sims24: See comment on poolstatus 

Thanks

network/f5/bigip/snmp/mode/poolstatus.pm
cgagnaire: Not that odd. The names are built from oid indexes that are ASCII translation of the names. But there always is a random charactere in first position.

Sims24: roger that, just wonder why this char is popping ... let's merge to let everybody get access to this fix

cgagnaire: Yea I know, hope its not a bug from my f5 box...

Sims24: Need to test it against our snmp simulator on friday. we got one walk from other user ;)

cgagnaire: Already tested, and there is no pools or virtualservers :/

PR #933, opened 2018-03-30
Fix stormshield hanodes mode

cgagnaire:
Fix dead nodes value catching and filter management so we just retreive what we need.
Add role of node.

### DIFF
--- a/network/stormshield/snmp/mode/hanodes.pm
+++ b/network/stormshield/snmp/mode/hanodes.pm
@@ -27,19 +27,27 @@ use warnings;
 
 my $instance_mode;
 
+sub custom_node_perfdata {
+    my ($self, %options) = @_;
+    
+    $self->{output}->perfdata_add(label => 'dead_nodes',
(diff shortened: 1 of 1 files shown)


network/stormshield/snmp/mode/hanodes.pm
```diff
@@ -152,15 +158,15 @@ sub new {
...
+                                "critical-state:s"      => { name => 'critical_state', default => '%{state} =~ /offline/i' },
                                 });
     return $self;
 }
 
 sub change_macros {
     my ($self, %options) = @_;
 
-    foreach (('warning_state', 'critical_state')) {
+    foreach ('warning_state', 'critical_state') {
```
Sims24: @cgagnaire think that double parenthesis is here for backward compatibility. Did you test this on CentOS 6 perl version ?

Sims24: Ok just test and it works. Do not know why Quentin used to write it like this as it's not a know convention, no reason so let merge this one.

PR #938, opened 2018-04-04
Convert blocked processes counters

omercier:
Hope everything is ok for my first serious PR on plugins...


Sims24:
Need your toughts on it @cgagnaire please give us your point of view.

Thanks

PR #974, opened 2018-04-30
enhance http lib to handle hidden basic authent

cgagnaire:
Kind of roll back https://github.com/centreon/centreon-plugins/pull/860 to get back transparent behaviour we had before when it comes to basic authentication.
By adding --basic option, we can handle hidden basic authentication and also prevent webserver to respond 401 errors for basic authentication.
Digest and NTLM authentication still don't need any options.


Sims24:
thanks @cgagnaire do we need to change the help and options on other modes using basic authentication ? (e.g apache server status mode)

cgagnaire:
Yes we do, it will come on another PR.

Sims24:
ok at least : 

nginx
apache 
tomcat webmanager

PR #979, opened 2018-05-03
 Using a file a JSON source did not work (wrong file handler) 

omercier:


Sims24:
hi @omercier thanks for the PR, could you give an environment or how to we can reproduce the bug. Seems odd it doesn't work

omercier:
Hi @Sims24 the material I have to reproduce the bug contains customer login/password so I can't give it to you here, but I sure can show you how I do it.

cgagnaire:
Thanks!

Issue #79, opened 2015-06-23
os::linux::snmp::plugin - mode storage and free space


cmeschin:
Bonjour,
l'utilisation de l'argument --free permet de travailler sur les espaces restants mais les metric remontés sont de type "free", le metric "used" n''apparait plus.
Cette utilisation n'est donc pas compatible avec Centreon BI et les rapports de capacité.

Est-il possible d'ajouter le metric "used" en complément du metric free pour ce mode de fonctionnement?
De plus, est-il possible de pouvoir spécifier les tailles disques dans des unités humainement lisible (M ou MB, G ou GB, T ou TB) plutôt qu'en B uniquement? le plugin ferait alors lui-même la conversion en B pour effectuer les calculs et remonter les metric corrects.
Cette gestion d'unité serait appréciable sur l'ensemble des modes gérant une taille en Bytes.

cordialement,

----- version anglaise ------
Hello, the use of the argument --free allows to work on the remaining spaces but the gone back up metric is of type "free", the metric "used" not appears more.
This use is not thus compatible with Centreon BI and the reports of capacity.

Is it possible to add the metric " used " as a supplement to the metric free for this way of functioning?
Furthermore, it is possible to be able to specify the sizes disks/memory in units humanely readable (M or MB, G or GB, T or TB) rather than B only? The plugin would make then itself the conversion B to make the calculations and go back up(raise) the correct metric.
This management of unity(unit) would be considerable on all the modes managing a size in Bytes.

Cordially,

garnier-quentin:
Thanks for the ticket. But could you write it in English please ? :)

garnier-quentin:
It could be done to specify the unit (B, MB, GB). But it's more complex to add used also when you check with free value.

cmeschin:
The plugin return the total size. So used=total-free, no?

garnier-quentin:
Not so easy because you have thresholds and thresholds are on free value.

cmeschin:
In this case, the used metric in for compatibility on Centreon BI. Thresholds are required?
so warning_used=total-warning_free and same for critical.

garnier-quentin:
Need to think about it. But it's a main change :)

cmeschin:
ok thank you.

cmeschin:
a last request: the same change must be made at least on windows plugin ;)

florian-asche:
I think, we need one option for input and one for output.

[...]

PR #995, opened 2018-05-28
Patch fix https://github.com/centreon/centreon-plugins/issues/79

lkco:
Adding feature : display "used" perfdata when you use "--free" option for os::linux::snmp::plugin storage check

Please add my feature request


Sims24:
Hi, 

Thanks for the proposed fix, as mentionned in the linked issue we need to think about it to keep thresholds displayed and correct in any case.

We will review and try to find a global solution to adress the feature of "free" and "used" displayed together.

lkco:
why don't you accept my pull request ? or do you have another solution?

Sims24:
Hi, 

Please see my comment below and read the thread you're commit is linked to.

Need to think about it but it's a main change.

PR #1000, opened 2018-05-30
Cannot find physical memory informations fix #994

moix:
hi there @maksoy74, I have same scenario, windows servers can reply with either _Physical memory_ or _Physical RAM_
Didn't have time before to create a pr with the fix but now that I saw your issue here it is. fixes #994 
Hope it helps, thanks

PR #1001, opened 2018-05-30
snmp-windows-memory | Include Physical RAM in calling memory OID

maksoy74:
Windows sometimes is using Physical RAM instead of Physical memory.

This small correction would allow to call either one and assign the OID


Sims24:
Already fixed with https://github.com/centreon/centreon-plugins/pull/1000 

Thanks for your help :!

PR #1020, opened 2018-06-18
Linux os

kaileu:
Adding 2 Linux Checks.
The First is for Ubuntu/Debian derived OS which uses apt to update.
The 2nd is a Script to Check Linux Filesystems if there are mounted as read-only. (e.g. if the Storage has Problems in Virtualized Environments the Filesystem mount itself als read-only.)

### DIFF
--- a/os/linux/local/mode/mountro.pm
+++ b/os/linux/local/mode/mountro.pm
@@ -1,0 +1,203 @@
+#
+# Copyright 2018 Centreon (http://www.centreon.com/)
+#
+# Centreon is a full-fledged industry-strength solution that meets
+# the needs in IT infrastructure and application monitoring for
+# service performance.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


Sims24:
Hello, 

Thanks that's really interesting, I would prefer to keep os/linux as the standard checks for all distribtions, and create a dedicated directory for debian, suse, centos etc ... relying on common checks and providing some extra checks specific to each distrib. 

May ou adapt you're PR regarding to this ? Also, please check the indentation of the option declaration. 

Regards,

kaileu:
I Split the PR to reflect the requested changes to split OS specific checks to seperate Files

Sims24:
Ok thanks, I'm closing this one then

kaileu:
@Sims24 Ther is one check left that checks if there are read-only Filesystems.
This one is equal on all Linux Systems maybe it was overlooked because for this one i dodn't make another PR

Issue #991, opened 2018-05-22
apps::protocols::x509::plugin : check subject + expiration


Rico29:
using apps::protocols::x509::plugin  --mode validity, it would be great to have a validity-mode which can check both subject and expiration date (=> check that the certificate is valid).

Actually you have to chose between issuer, subject or expiration

Rico29:
hello, any hint about this ticket ?

cgagnaire:
Implemented here : https://github.com/centreon/centreon-plugins/pull/1306

PR #1035, opened 2018-07-04
add support for validity-mode expiration-and-subject

Rico29:
#991

### DIFF
--- a/apps/protocols/x509/mode/validity.pm
+++ b/apps/protocols/x509/mode/validity.pm
@@ -1,5 +1,5 @@
 #
-# Copyright 2017 Centreon (http://www.centreon.com/)
+# Copyright 2018 Centreon (http://www.centreon.com/)
 #
 # Centreon is a full-fledged industry-strength solution that meets
 # the needs in IT infrastructure and application monitoring for
@@ -71,8 +71,8 @@ sub check_options {
         $self->{output}->add_option_msg(short_msg => "Please set the port option");
         $self->{output}->option_exit();
     }
-    if (!defined($self->{option_results}->{validity_mode}) || $self->{option_results}->{validity_mode} !~ /^expiration|subject|issuer$/) {
-        $self->{output}->add_option_msg(short_msg => "Please set the validity-mode option (issuer, subject or expiration)");
+    if (!defined($self->{option_results}->{validity_mode}) || $self->{option_results}->{validity_mode} !~ /^expiration|subject|issuer|expiration-and-subject$/) {
+        $self->{output}->add_option_msg(short_msg => "Please set the validity-mode option (issuer, subject, expiration or expiration-and-subject)");
         $self->{output}->option_exit();
     }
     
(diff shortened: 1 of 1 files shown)


Rico29:
I'm not a perl expert, so I guess it's posssible to put repetitive stuff in subs...

Rico29:
can someone take a look at this please ?

Sims24:
We are going to check you PR, thanks for your contribution. 

Do you have any public site where we can challenge the plugin against ?

Rico29:
Hi Simon
Sorry, you can challenge it on any https enabled website. I don't actually have a website with expired cert or wrong name.

apps/protocols/x509/mode/validity.pm
```diff
@@ -156,6 +156,44 @@ sub run {
...
+                $name = inet_ntoa($name);
+            }
+            if ($name =~ /$self->{option_results}->{subjectname}/mi) {
+                push @subject_matched, $name;
+            } else {
+                $self->{output}->output_add(long_msg => sprintf("Subject Name '%s' is also present in Certificate", $name), debug => 1);
+            }
+        }
+
+        if (@subject_matched == 0) {
```
Sims24: If I understand this one, you check the subject and if it's not OK then you do not try to check expiration date right ?

Sims24:
@Rico29 sounds like a really good idea to merge some checks. 

Would you think that performing every check and filter output is an option for you ? 

I guess we need a deep refactoring on this, in the meantime except the comment above (not a huge fan of skipping checks because other failed ...) I think we can merge this one and we'll refactor it in the future. Just need to check with @cgagnaire if he agrees with me

Rico29:
I totally agree with that ! I'll try to rewrite something next week

Sims24:
Any news here ? Let us know if we can help. 

Rgds

CPbN:
Really nice addidion @cgagnaire, thank you !
One comment perhaps, what about adding a `--timeout` option ?
Thank you :+1:

cgagnaire:
I've just added it (https://github.com/centreon/centreon-plugins/commit/3ece41a4245936ad8773c5ac68c62f7fcb1e3987).
Note that this option will work only with IO::Socket::SSL >= 1.984.
The mode itself will work only with Net::SSLeay >= 1.42.
I've tested latest version of IO::Socket::SSL and Net::SSLeay on both CentOS 6 and 7 and it works well.
We will push those versions on our repos.

cgagnaire:
Anyway, thanks @Rico29 for you work ;)

CPbN:
Many thanks @cgagnaire & @Rico29 :+1:

CPbN:
> The mode itself will work only with Net::SSLeay >= 1.42.
I've tested latest version of IO::Socket::SSL and Net::SSLeay on both CentOS 6 and 7 and it works well.

Strangely enough I face the following issue with CenOS 7 Perl-IO-Socket-SSL package, version 1.94-7 :
`SSL error: SSL wants a read first`

No issue with last version from CPAN, 2.060 (I've then fatpacked it).

> We will push those versions on our repos.

Would then be nice yes :+1:

Rico29:
hello @cgagnaire 
Any news about 
> We will push those versions on our repos.

Regards,
Cédric

Rico29:
@cgagnaire ?

CPbN:
> I've just added it ([3ece41a](https://github.com/centreon/centreon-plugins/commit/3ece41a4245936ad8773c5ac68c62f7fcb1e3987)).
> Note that this option will work only with IO::Socket::SSL >= 1.984.
> The mode itself will work only with Net::SSLeay >= 1.42.
> I've tested latest version of IO::Socket::SSL and Net::SSLeay on both CentOS 6 and 7 and it works well.
> We will push those versions on our repos.

Hi @cgagnaire,
Some news regarding this ?
Many thanks :+1:

cgagnaire:
Hi @Rico29 & @CPbN,
As @garnier-quentin said here https://github.com/centreon/centreon-plugins/issues/1414, we will not update it for now. There is possibility to break other modules.

PR #1036, opened 2018-07-04
Filesystem Check

kaileu:
Script to Check Linux Filesystems if there are mounted as read-only. (e.g. if the Storage has Problems in Virtualized Environments the Filesystem mount itself als read-only.)
https://github.com/centreon/centreon-plugins/pull/1020


Sims24:
We merge this one to work on it, few modifications still needed ;)

PR #1040, opened 2018-07-09
ntp daemon check (ntpd and chronyd)

kaileu:
Checks the NTP Daemon for valid NTP-Server
Supported commands ntpq and chronyc


Sims24:
Looks promising, thanks for posting your work. 

Wonder if it would be relevant to make a dedicated mode for both cronyc and ntp instead of manage the two possibilities within one mode. What do you think ? 

Think we probably have to refactor it with the counter template class to make the code easier to maintain and more readable.  Could you take a look at it ? 

Btw, default value for options are not a good practices as some people just look for metrology and that thresholds cannot be the same in every situations.

kaileu:
1. The decision to combine both services comes with idea to have one service to check ntp-server.
As there are only 2 real ntp-server on the market (ntpd and cronyc) i decided to combine both in one mode.

2. I am not sure what you mean with the counter template class. Could you point me to a hint or example then i will take a look at it.

3. If you want i can delete the default options.

Sims24:
1. Ok sounds good, thanks for your insights ;

2. Take a look at https://github.com/centreon/centreon-plugins/blob/master/centreon/plugins/templates/counter.pm this is the library we're using to developp quicker now.  Lot of example around but here is one : https://github.com/centreon/centreon-plugins/blob/c790526a4bc331c73fb87b1a4cc7713046ec8236/database/oracle/mode/eventwaitsusage.pm . manage_selection is the new run, your commands, monitoring logic etc should go there. At the end you pass value to your counter object(s) declared through the set_counter function. Give it a try ;) 

3. Yes I would prefer. Anyway if we refactor as mentionned in poin 2, it will be the case

garnier-quentin:
Sorry for the delay. Could you test it to know if it's ok for you ?

kaileu:
Thanks for your effort.
The Code seems pretty good to me there is only one Problem.
The Plugin only works for chrony 3 or greater because the -c switch is only available at this version.
So its working for Ubuntu 18.04 or greater but not on Ubuntu 16.04.
Debian Buster/Sid is working and Debian 8 is not working.

So this restriction has to keep in mind.

garnier-quentin:
Ok. I didn't know that. I will correct it.

garnier-quentin:
Could you test with the last version ? (should be ok)

kaileu:
The latest Version looks good.
I ran Plugin since yesterday and it looks good. :-)
Thanks!!

garnier-quentin:
Great!

Issue #936, opened 2018-04-04
[database::oracle::plugin] --mode='sql-string' 


bmareau:
Hi,

There is a problem with sql-string mode of oracle plugin.
When this command is used we have an error : 

```
/usr/lib/centreon/plugins/centreon_oracle.pl --plugin=database::oracle::plugin --hostname='xx.xx.xx.xx' --port='xxxx' --sid='SID' --username='user' --password='pws' --mode='sql-string' --sql-statement="select 'Texte' from dual" --datasource Oracle:xxxxxxxxxx.LOCAL/XXXXX06             

Use of uninitialized value in hash element at /usr/lib/centreon/plugins/centreon_oracle.pl line 420.                                                         
Use of uninitialized value in hash element at /usr/lib/centreon/plugins/centreon_oracle.pl line 420.                                                         
Use of uninitialized value in concatenation (.) or string at /usr/lib/centreon/plugins/centreon_oracle.pl line 420.                                          
OK: string : skipped (no value(s)) |   
```

This would be related to value-column and key-column.

Best Regards

Sims24:
Hi, 

It would be interesting to have the result of the query to be able to find the rott cause but sounds like a bug :) 

Thanks

Sims24:
We found a workaround with the fix below .. It ensures compatibility with dual table and Oracle. 

You can fix this by adding a dual_table option declaration and the code snippet below (after the fetchrow_hashref) : 

```
        if (defined($self->{option_results}->{dual_table})) {
            $row->{$self->{option_results}->{value_column}} = delete $row->{keys %{$row}};
            for (keys %{$row}) {
                $row->{$self->{option_results}->{value_column}} = $row->{$_};
            }
        }
```

Need to further discuss with guys to know how we can achieve this more properly

Giacomoni-GRF:
Hi,
Can you explain where we add these code on the file ? It's centreon_oracle.pl ?

Thanks .

Best regards,

PR #1075, opened 2018-07-23
enh sql sqlstring for dual tables

cgagnaire:


cgagnaire:
Fixes https://github.com/centreon/centreon-plugins/issues/936

PR #1080, opened 2018-07-24
enh h3c hardware mode to allow count thresholds

cgagnaire:
- Fix $self->{components_module}
- Add --short-name to make it more readable


network/h3c/snmp/mode/hardware.pm
```diff
@@ -171,6 +173,11 @@ sub new {
 sub check_options {
     my ($self, %options) = @_;
     $self->SUPER::check_options(%options);
+
+    $self->{short_name} = 0;
+    if (defined($self->{option_results}->{short_name})) {
```
Sims24: Could you please use ternary operator here ?

Issue #1088, opened 2018-08-03
apps::activedirectory::local::mode::dfsrbacklog add new option


lpinsivy:
On my MS Active Directory, if a run the following MS DOS command I receive:

```
C:\>dfsrdiag backlog /SendingMember:My_Sender /RGName:Public /RFName:Public
[ERROR] Replication group <public> not found Err: -2147217406 (0x80041002)
```

But If i run this command it's working:
```
C:\>dfsrdiag backlog /sendingmember:My_Sender /receivingmember:My_Receiver /RGname:Public /RFname:Public
No Backlog - member <My_Receiver> is in sync with partner <My_Sender>
```

Can you add in the plugin the option **--receiving-member** ?

cgagnaire:
Closed by https://github.com/centreon/centreon-plugins/pull/1089

PR #1089, opened 2018-08-03
add receiving-member option

lkappel:
add receiving-member option

issue #1088


Sims24:
hi @lkappel thanks, 

Cant't accept it. Option should not be mandatory, that's to handle some specific case and should not be the default as it will break all existing configurations.

cgagnaire:
Moreover, you don't use it :)

cgagnaire:
Can you just put a "$self->{option_results}->{command_options} .= '/ReceivingMember:' . $self->{option_results}->{receiving_member} . ' ' if defined ..."

PR #1091, opened 2018-08-03
enh mssql plugin database, add list-database mode

cgagnaire:
- Enhance databases-size mode, metrics will be "used" instead of "db\_\<dbname\>\_used", and "used\_\<dbname\>" instead of "db\_\<dbname\>\_used\_\<dbname\>",
- Remove low case on output,
- Add list-databases mode.


Sims24:
ok, need to explicitily identify this as a breaking change in the next tag

PR #1100, opened 2018-08-19
Add per-core CPU thresholds

### LINKED ISSUE #339: [check_centreon_snmp_cpu] cpu threshold per core?
is possible to define cpu threshold per core?
the cpu plugin only define threshold for the average cpu.

Thx

CPbN:
Hi,

This PR allows to apply thresholds on every core of CPUs, rather than on the overall CPU.
It then adds a new `percore` option.

And thus closes #339.

Thank you 👍

### DIFF
--- a/snmp_standard/mode/cpu.pm
+++ b/snmp_standard/mode/cpu.pm
@@ -21,6 +21,7 @@
 package snmp_standard::mode::cpu;
 
 use base qw(centreon::plugins::mode);
+use List::Util qw[min max];
 
 use strict;
 use warnings;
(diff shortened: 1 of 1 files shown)


Sims24:
Thanks, we will think about it but as usual when bringing enhancement about an old code, we would prefer a refactoring including counter templates.

CPbN:
Thank you @Sims24 for your feedback :+1: 
Yes this is more a fix to be able to detect a core stuck @100% rather than a complete plugin rewrite.

PR #1124, opened 2018-09-05
Add an autoreduce option to adapt UDP packet size

CPbN:
Hi,

Here is a PR to add a new `--autoreduce` SNMP option.
If a SNMP request fails because it was too big, `--autoreduce` will automatically retry a smaller request.

Even if we use for example `--maxrepetitions=20 --subsetleef=20` (which are fairly low values), Centreon Plugins may fail from time to time, depending on what is running on the requested device.
`--autoreduce` then helps here.

Thank you 👍

### DIFF
--- a/centreon/plugins/snmp.pm
+++ b/centreon/plugins/snmp.pm
@@ -53,6 +53,7 @@ sub new {
                   "maxrepetitions:s"          => { name => 'maxrepetitions', default => 50 },
                   "subsetleef:s"              => { name => 'subsetleef', default => 50 },
                   "subsettable:s"             => { name => 'subsettable', default => 100 },
+                  "autoreduce"                => { name => 'autoreduce' },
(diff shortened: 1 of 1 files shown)


Sims24:
Hi, 

Thanks for your help but we do not want to introduce usage of some "GOTO"/"RETRY" in our code. It should be avoided as much as possible. It's not mandatory here, we could handle this with classic loops.

We will not modify the snmp library unless we face more and more issues like the one you described.

CPbN:
_I pushed [a commit](https://github.com/CPbN/centreon-plugins/commit/e1d66e8b71c595a12b99efd414d83a90e1983058) to get rid of `goto` instructions, but unfortunately did not come up to this PR, as it is closed._

CPbN:
Hi,
Any thoughts ?
Thank you again !

PR #1129, opened 2018-09-06
Add a plugin for checking a domain name expiration date.

sboulianne:
Add a plugin for checking a domain name expiration date.
The plugin check_domain do that job but it cans be better.

https://github.com/glensc/monitoring-plugin-check_domain


Sims24:
Hi @sboulianne 

Please open an Issue rather than a Pull request for this kind of wish. 

Thanks

PR #1150, opened 2018-09-21
Add UDP rtt to Uptime mode

CPbN:
Hi,

This PR adds _round-trip time_ (RTT) info to SNMP _Uptime_ mode.
Sometimes, we can't use ICMP ping, due to remote device / firewall rules.
In this situation, one could still want to be able to build & display RTT graphs.
This PR then adds to the _Uptime_ mode the duration needed to get the Uptime OID, which is a tiny operation, thus can be considered as the _round-trip time_ (difference is less than 0.5ms in my tests).

Thank you :+1:

### DIFF
--- a/snmp_standard/mode/uptime.pm
+++ b/snmp_standard/mode/uptime.pm
@@ -27,7 +27,7 @@ use warnings;
 use POSIX;
 use centreon::plugins::misc;
 use centreon::plugins::statefile;
-use Time::HiRes qw(time);
+use Time::HiRes qw(time gettimeofday tv_interval);
 
 sub new {
     my ($class, %options) = @_;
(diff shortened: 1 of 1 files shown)


Sims24:
Hi, 

Sorry looks too specific and I do not agree with your statement that it's close to an rtt measurement. 

I guess that SNMP related things might have an impact on the result (e.g overloaded or buggy snmp agent) and globally this is not enough accurate.

CPbN:
Thank you for your answer @Sims24.
Perhaps we could name the data `snmprtt` instead of `rtt` ?

Think about hosts you can't ping, but you would like to graph & monitor "response time" value.
More than the absolute value itself and its accuracy, the variation is important.
I then think it's quite relevant.

perhaps we could make it a non-default option, so that one could enable it if needed ?

Many thanks !

cgagnaire:
You may want to try the apps::protocols::tcp::plugin plugin with the response-time mode.

CPbN:
Thank you @cgagnaire for your proposal.
I tested it before writing this PR, but unfortunately, it may not be suitable in some situations :-/

PR #1153, opened 2018-09-25
Add number of partitions in diskusage OK msg

CPbN:
Hi,

This PR adds number of seen partitions in SNMP standard diskusage plugin.

Thank you :+1:

### DIFF
--- a/snmp_standard/mode/diskusage.pm
+++ b/snmp_standard/mode/diskusage.pm
@@ -158,7 +158,7 @@ sub run {
 
     if (!defined($self->{option_results}->{diskpath}) || defined($self->{option_results}->{use_regexp})) {
         $self->{output}->output_add(severity => 'OK',
-                                    short_msg => 'All partitions are ok.');
+                                    short_msg => 'All '.$num_disk_check.' partitions are ok.');
     } elsif ($num_disk_check == 0) {
         $self->{output}->output_add(severity => 'OK',
                                     short_msg => 'No usage for partition.');


CPbN:
This allows to quickly see, when using for example a _regex_ to select the filesystems to monitor, that the number of monitored filesystems in place is correct.

CPbN:
Let's merge this tiny PR please ?
Many thx :+1:

cgagnaire:
Hi @CPbN,
I just refactored the mode so it uses the template counter.
Can you test this branch cgagnaire:enh-snmp-standard-diskusage-mode ?
Thanks!
(for your case, set --filter-counters='usage|count')

CPbN:
Hi @cgagnaire,
Thank you very much for your rework !

I then tested #1233, `--filter-counters=count` is interesting as it allows to verify that all expected FS are mounted.
(note that `--filter-counters` option is missing in the `--help` section, perhaps you should add it ?)

However, I would have liked the number of partitions to be displayed in `usage` mode, as proposed in this PR.
[Here in your PR #1233](https://github.com/centreon/centreon-plugins/pull/1233/files#diff-7eb447d6b0eb5bd1649ee71cc0956074R113), let's then add the number of partitions in the _multiple message_ :
`All X partitions are ok`

Many thanks :+1:

CPbN:
Hi @cgagnaire,
Thank you very much for your rework !

I then tested #1233, `--filter-counters=count` is interesting as it allows to verify that all expected FS are mounted.
(note that `--filter-counters` option is missing in the `--help` section, perhaps you should add it ?)

However, I would have liked the number of partitions to be displayed in `usage` mode, as proposed in this PR. It's just a piece of info, it helps, when deploying monitoring, or looking at service status, to know that the _regexp_ used is correct. For example, without this info, I was monitoring some Docker internal FS which did not make sense. Once I made this PR, I noticed that the number of monitored FS was clearly too high, and then updated my `regexp`.
[Here in your PR #1233](https://github.com/centreon/centreon-plugins/pull/1233/files#diff-7eb447d6b0eb5bd1649ee71cc0956074R113), let's then add the number of partitions in the _multiple message_ :
`All X partitions are ok`

Many thanks :+1:

cgagnaire:
Hi @CPbN,
I think this commit meet your need as you'll see the number of partitions.
As you can see, its a dedicated counter. This way you can even set thresholds on it.
We will not mix it up with the multiple counter.
And frankly, what's the difference between `OK: Partitions count : 11 - All partitions are ok` and `All 11 partitions are ok `

CPbN:
The problem is that if I use `--filter-counters=count` to have the number of partitions monitored in the message, I can't monitor usage anymore.
At least I tested and it does not seem to work, as we have to choose between `--filter-counters=usage` (default) and `--filter-counters=count`.

cgagnaire:
No, as I said, use --filter-counters='usage|count' and you'll have both :)

CPbN:
Perfect, I did not understand it like this :)
Just tested, it works :+1:
Thank you @cgagnaire, let's then close this !

cgagnaire:
OK! Thanks for using the project!

PR #1154, opened 2018-09-26
Fix azure plugin

cgagnaire:


cgagnaire:
Closest to https://azure.microsoft.com/en-us/services/

PR #1230, opened 2018-11-20
fix cisco callmanager phone name oid

cgagnaire:
Use name instead of description as its not always defined.


cgagnaire:
<3 https://github.com/centreon/centreon-plugins/pull/1231

PR #1233, opened 2018-11-20
enh snmp_standard diskusage mode

cgagnaire:
Refactor snmp_standard plugin diskusage mode to use template counter.
Add a global counter to count the number of partitions for defined filters.


cgagnaire:
Breaks --warning and --critical thresholds, replaced by --warning-usage and --critical-usage.

PR #1255, opened 2018-12-05
fix(doc) typo in help

tseegers:


cgagnaire:
Hi @tseegers,
It's not quite that, there is a typo but it should be : 
Example: 0,OK,File xxx exist#1,CRITICAL,File xxx not exist#3,UNKNOWN,Command problem
I'll change it for you.
Thanks!

cgagnaire:
Well no. If the numeric value is not defined, its default so...
The example is good.

PR #1259, opened 2018-12-06
Centreon Connector Perl compatibility

ganoze:
Add __END__ at the end of centreon_plugins.pl


cgagnaire:
(doing some tests...)

ganoze:
Useless indeed.

Issue #880, opened 2018-02-07
Use of uninitialized value $option in sprintf at /root/Centreon Install/centreon-plugins/storage/netapp/snmp/mode/volumeoptions.pm line 109.


Saix00:
Hi,

We're configuring a snmp for our NetApps systems ang with the mode volumeoptions for the NetApp plugin we're seeing the following errors:

`[root@Centreon-Central` centreon-plugins]# perl centreon_plugins.pl  --plugin=storage::netapp::snmp::plugin --mode=volumeoptions --hostname *** --snmp-community *** --snmp-version 2
Use of uninitialized value $option in sprintf at /root/Centreon Install/centreon-plugins/storage/netapp/snmp/mode/volumeoptions.pm line 109.
Use of uninitialized value $option in sprintf at /root/Centreon Install/centreon-plugins/storage/netapp/snmp/mode/volumeoptions.pm line 109.
Use of uninitialized value $option in sprintf at /root/Centreon Install/centreon-plugins/storage/netapp/snmp/mode/volumeoptions.pm line 109.
Use of uninitialized value $option in sprintf at /root/Centreon Install/centreon-plugins/storage/netapp/snmp/mode/volumeoptions.pm line 109.
Use of uninitialized value $option in sprintf at /root/Centreon Install/centreon-plugins/storage/netapp/snmp/mode/volumeoptions.pm line 109.
Use of uninitialized value $option in sprintf at /root/Centreon Install/centreon-plugins/storage/netapp/snmp/mode/volumeoptions.pm line 109.
Use of uninitialized value $option in sprintf at /root/Centreon Install/centreon-plugins/storage/netapp/snmp/mode/volumeoptions.pm line 109.
Use of uninitialized value $option in sprintf at /root/Centreon Install/centreon-plugins/storage/netapp/snmp/mode/volumeoptions.pm line 109.
Use of uninitialized value $option in sprintf at /root/Centreon Install/centreon-plugins/storage/netapp/snmp/mode/volumeoptions.pm line 109.
Use of uninitialized value $option in sprintf at /root/Centreon Install/centreon-plugins/storage/netapp/snmp/mode/volumeoptions.pm line 109.
Use of uninitialized value $option in sprintf at /root/Centreon Install/centreon-plugins/storage/netapp/snmp/mode/volumeoptions.pm line 109.
[...]

PR #1316, opened 2019-01-09
Improve snmp-autoreduce in case device does not reply tooBig

CPbN:
Hi,

I faced issue described in #880 :
`--plugin=storage::netapp::snmp::plugin --mode=volumeoptions` asks for info for let's say 25 volumes.
The Netapp device replies but only for the X first volumes (certainly until a fixed packet size is reached, around 4000 bytes as per my tests).
There is no SNMP error : `$self->{session}->get($vb)` in `get_leef()` does not return error.
But then some elements are missing in the answer, which then leads to errors (and undefined values) described in #880.

I'm then quite surprised that the Netapp device does not reply with the `tooBig` error message.
This PR then fixes this when `--snmp-autoreduce` is used.

I wonder if we would have to apply the same fix into `get_table()` and `get_multiple_table()`.
According to their loop through `@$vb` I would say no as `${$entry}[2]` is checked there without testing `@{$entry}` length.
But perhaps this is an issue...

Thank you :+1:

### DIFF
--- a/centreon/plugins/snmp.pm
+++ b/centreon/plugins/snmp.pm
@@ -324,6 +324,11 @@ sub get_leef {
             return undef;
         }
 
+        # Some devices could not send "tooBig" ErrorNum ?
+        if(($self->{snmp_autoreduce} == 1) && scalar($vb) && (scalar(@{@$vb[-1]}) < 3)) {
+            next if ($self->autoreduce_leef(current => $entry) == 0);
+        }
+
         foreach my $entry (@$vb) {
             if ($#$entry < 3) {
                 # Can be snmpv1 not find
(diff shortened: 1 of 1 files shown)


CPbN:
In reality it's even worse.
Let's say you ask for 25 volumes.
And you have `--subsetleef=20`.
You'll get an answer for the 7 first volumes (UDP packet around 4000 bytes), and the 5 last ones (second SNMP request).
You won't get info for volumes 8 to 20 (part of the first SNMP request).

garnier-quentin:
It shouldn't work that way. Could you do a teamviewer together tomorrow ?

CPbN:
Sure !

garnier-quentin:
Fixed 3c48822071cecaf60a1357885e17fa5c86d442a4

CPbN:
Many thanks :+1:

PR #1325, opened 2019-01-10
Harden SNMP request even more

CPbN:
Hi,

This follows https://github.com/centreon/centreon-plugins/commit/3c48822071cecaf60a1357885e17fa5c86d442a4, hardening SNMP request one step further, we never know...

Thank you :+1:


garnier-quentin:
Why not!

PR #1333, opened 2019-01-12
x509, do not remove subject from alt_subjects

CPbN:
Hi,

This PR prevents removing `subject` from `alt_subjects` in certificate plugin.
This leads to errors where `alt_subjects` is used for example in a Centreon service template to simply check that certificates at least contain the expected domain.

Thank you :+1:


CPbN:
Many thanks @cgagnaire :+1:

PR #1336, opened 2019-01-13
HTTP expected-content, treat the string as single line

CPbN:
Hi,

This PR treats the HTTP expected-string as single line.
This then allows to do things like this :
`--expected-string='My nice site</title>.*</html>'`

Thank you :+1:


CPbN:
Perhaps it's better for the user to enforce his expected behavior using `(?sm)` at the beginning of his regex, instead of modifying the current plugin behavior.
I'll then close this for now, feel free to comment if needed.
Thx :+1:

PR #1338, opened 2019-01-14
Netapp, check volumes state

CPbN:
Hi,

This PR allows to check for volumes state (in addition to options).

Thank you :+1:


garnier-quentin:
I have changed the logic. You have new options :). Could you test it ?

PR #1340, opened 2019-01-14
Uptime, reduce chances to miss a reboot

CPbN:
Hi,

This PR reduces the chances to miss a reboot when uptime `--check-overload` is used.
I'm not sure why a +/- 50 seconds window would be required, which seems overkill.
A +/- 2.5 seconds window is IMO enough, and we should not miss any reboot then (as generally they last longer).

Many thanks :+1:

### DIFF
--- a/snmp_standard/mode/uptime.pm
+++ b/snmp_standard/mode/uptime.pm
@@ -69,7 +69,7 @@ sub check_overload {
     return $options{timeticks} if (!defined($self->{option_results}->{check_overload}));
     
     my $current_time = floor(time() * 100);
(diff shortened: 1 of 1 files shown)


garnier-quentin:
Because it can be timeout and other. So i prefer to let 50 seconds window. But we can add an option to set the window.

CPbN:
That would say that SNMP answer has taken more than 50 seconds to come back from the device (well this is the only reason I see).
I've pushed another commit to harden `--check-overload` / correct another possible issue.

CPbN:
I've just have submitted a new approach that should avoid missing reboots :)
It also solves one or two issues.
Thank you !

garnier-quentin:
I have added an option you can use.

PR #1342, opened 2019-01-14
Netapp, check snapmirrors state

CPbN:
Hi,

This PR improves Netapp snapmirrors check, checking their state in addition to their lag.

Thank you 👍

### DIFF
--- a/storage/netapp/snmp/mode/snapmirrorlag.pm
+++ b/storage/netapp/snmp/mode/snapmirrorlag.pm
@@ -114,7 +114,7 @@ sub run {
     $self->{snmp}->load(oids => [$oid_snapmirrorState, $oid_snapmirrorLag], instances => $self->{snapmirrors_id_selected});
     my $result = $self->{snmp}->get_leef(nothing_quit => 1);
 
-    my $destinations = 0;
+    my $targets = 0;
     foreach my $instance (sort @{$self->{snapmirrors_id_selected}}) {
         my $name = $self->{result_names}->{$oid_snapmirrorSrc . '.' . $instance};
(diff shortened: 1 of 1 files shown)


CPbN:
May I ask you to review this and merge if it's OK for you ?
Many thanks !

garnier-quentin:
Could you test it ?

CPbN:
Many thanks Quentin !
You did not kept the filter-out on source volumes ?
https://github.com/centreon/centreon-plugins/pull/1342/files#diff-002875028676178aa324eaa7afd23cc2R123
This allows to quickly show if all snap targets are on the same host, and allows to generate only one alert (from the target host) when there's an issue, instead of 2.
Thx :+1:

CPbN:
And of course I just tested the plugin, it works perfectly, thank you again :+1:

garnier-quentin:
It will work even in source. I don't really see the point. I have done a last change. You can test it a last time ?

CPbN:
Perfect, works as expected :+1: 
About source volumes, goal is to exclude them.
Think about a 2 nodes cluster, with the same monitoring on the 2 nodes.
You'll then monitor all snapmirrors twice.
Could be then interesting to exclude "source" volumes, so that :
- alerts will come from one node only ;
- you'll be able to know if all targets are on the same host (you expect 0 snapmirror on one of the 2 nodes).

garnier-quentin:
Can add an option to filter on state so.

CPbN:
Nice solution yes !

garnier-quentin:
Add filter-status option

CPbN:
Works perfectly, thank you @garnier-quentin !

PR #1348, opened 2019-01-17
Adds deduptvtl mode to quantum dxi plugin

tanguyvda:


cgagnaire:
Hi @tanguyvda,
Can you please use the latest implementation of common functions for the status thresholds (like here https://github.com/centreon/centreon-plugins/blob/master/storage/quantum/dxi/ssh/mode/dedupnas.pm) ?
Thanks!

PR #1356, opened 2019-01-22
added option 'poller' in pollerdelay.pm

da-friedl:
Hi,
we have a lot of Centreon poller in our infrastructure, also monitored via Centreon. We wanted to have the ability to check the delay of each individual poller.


garnier-quentin:
Thank you for the patch. We are going to refactoring the mode (with a filter on poller name).

garnier-quentin:
Done in the last version

PR #1378, opened 2019-02-05
Add support for fetching perfdatas and format, from the JSON

BenoitPoulet:
Hello,

We added support for fetching perfdatas and the message format from the JSON.
This extend the plugin, so we can check a JSON as if the JSON was a check result.

This can be used to check a service who is made in a asynchonous way, and who expose his result in JSON on a web server.

Example with this screenshot :
capture_json

The JSON contain information status. "Etat : OK", and some perfdata.

So with this command, i can check if i got "KO" as "Etat" and use the perfdata from the JSON :

$ ./centreon_plugins.pl --plugin=apps::protocols::http::plugin --mode=json-content --hostname="10.66.113.52" --urlpath='/monitoring.json' --lookup="$.Status.Etat" --header="Content-Type: application/json" --proto=https --critical-string "KO" --format-critical="Houston, we have a problem" --format-ok "All is fine" --timeout 10 --lookup-perfdatas="$.Perf.Datas" --filter-perfdata='^((?!count|time).)*$'
OK: All is fine | 'rta'=10.752ms;50.000;100.000;0; 'pl'=0%;20;40;; 'rtmax'=10.802ms;;;; 'rtmin'=10.709ms;;;; 'rym'=4;;;;

Regards

### DIFF
--- a/apps/protocols/http/mode/jsoncontent.pm
+++ b/apps/protocols/http/mode/jsoncontent.pm
@@ -77,8 +77,9 @@ sub new {
             "format-ok:s"             => { name => 'format_ok', default => '%{count} element(s) found' },
             "format-warning:s"        => { name => 'format_warning', default => '%{count} element(s) found' },
             "format-critical:s"       => { name => 'format_critical', default => '%{count} element(s) found' },
-            "format-unknown:s"       => { name => 'format_unknown', default => '%{count} element(s) found' },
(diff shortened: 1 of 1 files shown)


garnier-quentin:
Could you change the option: --lookup-perfdata-nagios ?

BenoitPoulet:
As requested, option name changed : lookup-perfdatas -> lookup-perfdatas-nagios

PR #1397, opened 2019-02-22
Small response-time snmp mode hardening

CPbN:
Small improvement @garnier-quentin after https://github.com/centreon/centreon-plugins/commit/8f33251a6bc6df4715669bc97036cd6666143ce3.

Thank you :+1:


garnier-quentin:
That option is useless. Otherwise, you use --snmp-autoreduce option. You use it ?

CPbN:
I set some SNMP parameters system-wide, and autoreduce is part of them.
One could also manually set it, so it's just to avoid possible issue with this, to be sure to try only one time, without retrying. Just to harden the test.

CPbN:
Thank you Quentin !

Issue #1427, opened 2019-03-19
[templates::counter] Compute max value


CPbN:
Hi,

Would be nice to compute max value for counter template.

Here's an example :
`--plugin=os::linux::snmp::plugin --mode=cpu` currently gives as perfdata :
`'total_cpu_avg'=5.00%;;;0;100 'cpu_0'=9.00%;;;0;100 'cpu_1'=1.00%;;;0;100`

Would be nice to add in perfdata :
`'total_cpu_max'=9.00%;;;0;100`

We would then be able to graph the highest used CPU, whatever it is.

Thank you 👍

garnier-quentin:
I see but i'm think it should be done by your monitoring software.

CPbN:
So directly by Centreon itself you mean ?

garnier-quentin:
Centreon or icinga or other systems :)

PR #1432, opened 2019-03-20
Standard CPU, add the max used core in perfdata

CPbN:
Hi,

This PR adds the max used CPU core stats in standard CPU mode perfdata output.

Thank you :+1:

### DIFF
--- a/snmp_standard/mode/cpu.pm
+++ b/snmp_standard/mode/cpu.pm
@@ -30,6 +30,7 @@ sub set_counters {
 
     $self->{maps_counters_type} = [
         { name => 'cpu_avg', type => 0, cb_prefix_output => 'prefix_cpu_avg_output' },
+        { name => 'cpu_max', type => 0 },
         { name => 'cpu_core', type => 1, cb_prefix_output => 'prefix_cpu_core_output' }
     ];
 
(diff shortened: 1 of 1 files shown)


garnier-quentin:
Sorry, but i should be done on your monitoring system (to get a curve with max values).

CPbN:
Grafana easily does this, but I'm not sure about Centreon itself, so this was the easiest way.
I'll then dig into Centreon graph possibilities.

garnier-quentin:
If centreon cannot, you can ask for it in centreon github.

CPbN:
Still @garnier-quentin, there is one major advantage computing max value at plugin level.
Think about a CPU with 24 cores.
You don't care about the 24 individual core usage values, but you only need the max value.
With max value computed at plugin level (and with `--filter-perfdata` to remove individual core values), you only store one data series (in SQL and RRD) instead of 24.
Which is a nice performance and storage improvement.

garnier-quentin:
It depends what you want. With change-perfdata and extend-perfdata option, it cannot work. It will need another option.

CPbN:
I meant that with this PR we would get :
`'total_cpu_avg'=5.00%;;;; 'cpu_0'=9.00%;;;; ... 'cpu_23'=1.00%;;;; 'cpu_max'=9.00%;;;;`

So, using `--filter-data='total_cpu_avg|cpu_max'` we would easily get :
`'total_cpu_avg'=5.00%;;;; 'cpu_max'=9.00%;;;;`

Of course this PR is related to this CPU mode only, and as first thought in #1427, perhaps a more general option could make sense, if ever you think this makes sense :)

CPbN:
Thanks to #1442 :
` --filter-perfdata='^total_cpu_avg$|^core_cpu_max$'`
`--extend-perfdata-group=',core_cpu_max,max(cpu_[0-9+]),%'`
👍

PR #1445, opened 2019-03-26
Fix no-ha-status option

lordtaki:
'no_ha_status' or 'no_ha_state' to handle the option no-ha-status? :)
Let's use the first name in the code : 'no_ha_status'


garnier-quentin:
Thanks!

PR #1466, opened 2019-04-03
Add keepalived MIB support into VRRP mode

CPbN:
Hi,

Some devices do not support VRRP MIB, or at least do not fully support it...
This PR then adds KEEPALIVED MIB as an option to support VRRP mode, as we also found the needed OIDs there.

Many thanks :+1:


garnier-quentin:
Sorry. We are going to do a specific plugin for keepalived. Could you give us a snmpwalk ? :)

CPbN:
OK :)
Sure, I'll come back to you by mail with the snmpwalk.
Let's then close this.
Thx !

PR #1471, opened 2019-04-08
Update HTTP default message

CPbN:
Hi,

Due to the the wide range of tests HTTP expected-content mode is able to perform (content, value, return code, response time, size...), let's update the default OK short message to a more generic one.

Thank you :+1:


CPbN:
Thank you ;)

PR #1497, opened 2019-05-07
Fix variable error riverbed

jben94:
Hi,

Little patch for Riverbed Load.

Regards


Sims24:
Thanks for spotting this @jben94! 

I commited to add the new perfdata system also and avoiding merge commit.

PR #1499, opened 2019-05-07
Add new Vyatta plugin

CPbN:
Hi,

This PR adds a new plugin to check Vyatta.
Rather simple.
The `keepalived` mode is more or less a copy of the `vrrp` one, with other OIDs of course.

Thank you :+1:

### DIFF
--- a/network/vyatta/snmp/plugin.pm
+++ b/network/vyatta/snmp/plugin.pm
@@ -1,0 +1,52 @@
+#
+# Copyright 2019 Centreon (http://www.centreon.com/)
+#
+# Centreon is a full-fledged industry-strength solution that meets
+# the needs in IT infrastructure and application monitoring for
+# service performance.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+package network::vyatta::snmp::plugin;
+
+use strict;
+use warnings;
+use base qw(centreon::plugins::script_snmp);
+
+sub new {
+    my ($class, %options) = @_;
+    my $self = $class->SUPER::new(package => __PACKAGE__, %options);
+    bless $self, $class;
+
+    $self->{version} = '1.0';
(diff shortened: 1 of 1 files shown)


Sims24:
Hi @CPbN thanks for this one ! 

Just wonder if the keepalived mode wouldn't deserve a dedicated directory in apps/keepalived. 

The fact that it's a ha/lb framework and that OIDs are part of a specific keepalived-mib, somebody could probably be interested in using the plugin but not necesserarly with a Vyatta equipement. 

What do you think ?

Sims24:
Also, do you have the following OIDs populated for keepalived part ? 

.1.3.6.1.4.1.9586.100.5.3.3 (virtualServerTable)
and
.1.3.6.1.4.1.9586.100.5.3.4 (realServerTable)

CPbN:
Thank you @Sims24 for your feedback.

> Just wonder if the keepalived mode wouldn't deserve a dedicated directory in apps/keepalived.

Good question.
There was a first try for the keepalived mode in #1466, well, for the VRRP instances info from the keepalived MIB I should say.
As the mode does let's say exactly what vrrp mode already does, but using different OIDs, I put it into the same `/snmp_standard/` folder.
One interested in this mode but for a device different than a Vyatta could easily integrate it into the `plugin.pm` file of its device.
We could also create a new `/apps/keepalived/` folder, up to you :)

> Also, do you have the following OIDs populated for keepalived part ?

Unfortunately no, on the 10 Vyatta I have, only `.1.3.6.1.4.1.9586.100.5.1` and `.2` are available.

Thank you :+1:

Sims24:
>> There was a first try for the keepalived mode in #1466, well, for the VRRP instances info from the keepalived MIB I should say.

Ok so you need to create a dedicated plugin.pm and move the keepalived mode to a dedicated folder in apps/keepalived/snmp/ . 

>> As the mode does let's say exactly what vrrp mode already does, but using different OIDs, I put it into the same /snmp_standard/ folder.

snmp_standard is for all OIDs described by a RFC and should not contain anything related to a vendor_id, so please move it to the directory mentionned above. 

>> One interested in this mode but for a device different than a Vyatta could easily integrate it into the plugin.pm file of its device.

Yes ! 

Regards

CPbN:
Just modified as you requested @Sims24, I then created a dedicated Keepalived plugin in #1505.
Thank you :+1:

CPbN:
Well, as all remaining modes in this Vyatta plugin can be achieved using `os::linux::snmp::plugin`, I'm not sure it is really relevant...
In addition Vyatta is simply a Linux appliance...
Feel free then to tell me if we should rather close this...
Thank you !

Sims24:
Yeah your right as far as I understand it only supports standard MIBs (https://support.rackspace.com/how-to/vyatta-snmp-and-logging/) so we will not create a dedicated directory for it.

PR #1488, opened 2019-04-23
Cisco Stack - added option to ignore if ring is not redundant

da-friedl:
Hi,

The current release will exit the script if the stack ring is not redundant. We have some cases at our customer sites where the stack ring is not wanted to be redundant, but we want to know the all the information about the non-redundant stack (reason may be due to costs of a second cable, the potential risk is accepted by the customer).

Best regards

Daniel

PR #1503, opened 2019-05-09
added binary option to ignore if a Cisco Switch Stack is not redundant

da-friedl:
Hi Sims24,

thank you very much for your review in #1488; i cleaned up my code, made a branch, re-implemented the code with your suggestions and tested the code against some switches in our network. All working well now. 

Best regards

### DIFF
--- a/centreon/common/cisco/standard/snmp/mode/stack.pm
+++ b/centreon/common/cisco/standard/snmp/mode/stack.pm
@@ -186,6 +186,7 @@ sub new {
                                 {
                                   "warning-status:s"  => { name => 'warning_status', default => '' },
                                   "critical-status:s" => { name => 'critical_status', default => '%{state} !~ /ready/ && %{state} !~ /provisioned/' },
+				  "ignore-cswringredundant" => { name => 'ignore_cswringredundant'},
                                 });
 
     return $self;
@@ -229,7 +230,6 @@ my $oid_cswRingRedundant = '.1.3.6.1.4.1.9.9.500.1.1.3.0';
 sub manage_selection {
     my ($self, %options) = @_;
     $self->{snmp} = $options{snmp};
-
     $self->{global} = { waiting => 0, progressing => 0, added => 0, ready => 0, sdmMismatch => 0, 
         verMismatch => 0, featureMismatch => 0, newMasterInit => 0, provisioned => 0, 
         invalid => 0, removed => 0 };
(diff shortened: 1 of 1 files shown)


centreon/common/cisco/standard/snmp/mode/stack.pm
```diff
@@ -239,7 +239,11 @@ sub manage_selection {
 
     if ($redundant->{$oid_cswRingRedundant} != 1) {
         $self->{output}->add_option_msg(short_msg => "Stack ring is not redundant");
-        $self->{output}->option_exit();
+
```
Sims24: Just a last thing, please prefer keep option exit function call side by side.  One if is enough here : 

```
    if (($redundant->{$oid_cswRingRedundant}) != 1 && (!defined($self->{option_results}->{ignore_cswringredundant}))) {
        $self->{output}->add_option_msg(short_msg => "Stack ring is not redundant");
        $self->{output}->option_exit();
    }
```

da-friedl: Yes, i understand! 

First we thought it would be helpfull to print out that the "stack ring is not redundant" here - so we're able to use the search output-filter in the Centreon GUI. I will add your recommendation. 

Further i will think about a proper solution to get a warning-message/output/status if the stack ring is not redundant (with a binary option at the command line).

Sims24: Hum yeah you're right, could be interesting. 

We could imagine to set redundancy state as a counter, this way user can set Threshold like :  

--warning-status='%{stack} =~ /Not redundant'

and see the Redundancy state in the output ? 

Are you interested in such thing ? If so I could push the relevant evolution on your fork OR even better, you could find some inspiration here (https://github.com/centreon/centreon-plugins/blob/505de51edc697f30101dfd3774c1ef06b4e4ca40/network/fortinet/fortimanager/snmp/mode/devicestatus.pm#L37) 

Let me know but think your proposition is relevant and deserve to be implemented in this PR

garnier-quentin:
Could you provide me a snmpwalk of your cisco equipment ?
```
snmpwalk -ObentU -c public -v 2c IP .1.3.6.1.4.1.9.9.500.1 > cisco.snmpwalk
```

garnier-quentin:
You can use following option to ignore the stack status:
```
--critical-stack-status=''
```

PR #1510, opened 2019-05-15
Support Mikrotik PHY error counters

### LINKED ISSUE #1485: [network::mikrotik] Support PHY error counters
Hi,

Would be nice, as already done on other network devices (such as Cisco ones), to support Mikrotik interface error counters.

I've detailed them below, from the `1.3.6.1.4.1.14988.1.1.14` MIB group.

RX counters :
```
.33 : mtxrInterfaceStatsRxTooShort
.41 : mtxrInterfaceStatsRxTooLong
.45 : mtxrInterfaceStatsRxFCSError
.46 : mtxrInterfaceStatsRxAlignError
.47 : mtxrInterfaceStatsRxFragment
.48 : mtxrInterfaceStatsRxOverflow
.50 : mtxrInterfaceStatsRxUnknownOp
.51 : mtxrInterfaceStatsRxLengthError
.52 : mtxrInterfaceStatsRxCodeError
.53 : mtxrInterfaceStatsRxCarrierError
.54 : mtxrInterfaceStatsRxJabber
.55 : mtxrInterfaceStatsRxDrop
```

TX counters :
```
.63 : mtxrInterfaceStatsTxTooShort
.71 : mtxrInterfaceStatsTxTooLong
.75 : mtxrInterfaceStatsTxUnderrun
.76 : mtxrInterfaceStatsTxCollision
.77 : mtxrInterfaceStatsTxExcessiveCollision
.78 : mtxrInterfaceStatsTxMultipleCollision
.79 : mtxrInterfaceStatsTxSingleCollision
.80 : mtxrInterfaceStatsTxExcessiveDeferred
.81 : mtxrInterfaceStatsTxDeferred
.82 : mtxrInterfaceStatsTxLateCollision
.83 : mtxrInterfaceStatsTxTotalCollision
.85 : mtxrInterfaceStatsTxDrop
.86 : mtxrInterfaceStatsTxJabber
.87 : mtxrInterfaceStatsTxFCSError
.89 : mtxrInterfaceStatsTxFragment
```

Many counters, but Mikrotik has only one MIB file for all their different PHYs.
Some devices may then support some OIDs from this list, other devices some other OIDs...
It's then expected not to find some of these OIDs on some devices.

Attached is the MIB file, as well as a snmpwalk for such a device.
As stated above, this device may then not implement all the listed counters.

Thank you very much 👍

[mikrotik.mib.txt](https://github.com/centreon/centreon-plugins/files/3094716/mikrotik.mib.txt)
[mtxrInterfaceStats.txt](https://github.com/centreon/centreon-plugins/files/3094717/mtxrInterfaceStats.txt)

CPbN:
Hi,

This closes #1485, adding Mikrotik PHY error counters to interface mode.

Just one thing I'm not really sure about, the numeric prefix of each error counter in the `{maps_counters}->{int}` hash.
I then took OID + 100, feel free to tell me if it needs to be updated.

One limitation we see with this mode is that we miss a short way to define a threshold for all error counters at once.
Here we would need 31 x 2options to define warning and critical thresholds on all error counters.
If we set the same threshold on all counters, only 2 global options would be needed (one warning, one critical).
They would eventually be superseded by the specific ones, if used.
Another way would be to be able to define thresholds on `--extend-perfdata-group` option, but perhaps more difficult to achieve.

Thank you :+1:

### DIFF
--- a/network/mikrotik/snmp/mode/interfaces.pm
+++ b/network/mikrotik/snmp/mode/interfaces.pm
@@ -457,6 +457,7 @@ Can be: 'total-port', 'total-admin-up', 'total-admin-down', 'total-oper-up', 'to
 'in-traffic', 'out-traffic', 'in-error', 'in-discard', 'out-error', 'out-discard',
 'in-ucast' (%), 'in-bcast' (%), 'in-mcast' (%), 'out-ucast' (%), 'out-bcast' (%), 'out-mcast' (%),
 'speed' (b/s).
+
 And also: 'in-tooshort' (%), 'in-toolong' (%), 'in-fcserror' (%), 'in-alignerror' (%), 'in-fragment' (%),
 'in-overflow' (%), 'in-unknownop' (%), 'in-lengtherror' (%), 'in-codeerror' (%), 'in-carriererror' (%),
 'in-jabber' (%), 'in-drop' (%), 'out-tooshort' (%), 'out-toolong' (%), 'out-underrun' (%),
@@ -471,6 +472,7 @@ Can be: 'total-port', 'total-admin-up', 'total-admin-down', 'total-oper-up', 'to
 'in-traffic', 'out-traffic', 'in-error', 'in-discard', 'out-error', 'out-discard',
 'in-ucast' (%), 'in-bcast' (%), 'in-mcast' (%), 'out-ucast' (%), 'out-bcast' (%), 'out-mcast' (%),
 'speed' (b/s).
+
 And also: 'in-tooshort' (%), 'in-toolong' (%), 'in-fcserror' (%), 'in-alignerror' (%), 'in-fragment' (%),
 'in-overflow' (%), 'in-unknownop' (%), 'in-lengtherror' (%), 'in-codeerror' (%), 'in-carriererror' (%),
 'in-jabber' (%), 'in-drop' (%), 'out-tooshort' (%), 'out-toolong' (%), 'out-underrun' (%),
--- a/network/mikrotik/snmp/mode/interfaces.pm
+++ b/network/mikrotik/snmp/mode/interfaces.pm
@@ -1,0 +1,551 @@
+#
+# Copyright 2019 Centreon (http://www.centreon.com/)
+#
+# Centreon is a full-fledged industry-strength solution that meets
+# the needs in IT infrastructure and application monitoring for
+# service performance.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+package network::mikrotik::snmp::mode::interfaces;
+
+use base qw(snmp_standard::mode::interfaces);
+
+use strict;
+use warnings;
+
+sub set_oids_errors {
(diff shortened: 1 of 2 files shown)


garnier-quentin:
You need to update your PR.

CPbN:
Yes you're right, I'll do this ASAP.
One related question, do you plan to add `nlabel` / `new-perfdata` to interfaces mode ?

garnier-quentin:
It's done. You can check it

garnier-quentin:
I will do the update for you. Don't worry

CPbN:
Plugin updated to support counters template and new-perfdata.
I also now use loops which looks better.
As stated above, I call `$self->set_oids_errors();` from `set_counters_errors()` instead of from `load_errors()`.
I think you'll encounter this in other interface modes too.
Thx :+1:

garnier-quentin:
Why do you call it from set_counters_errors ? If you prefer :)

CPbN:
Because since you moved interfaces mode to counters template, `load_errors()` is called after `set_counters_errors()`.
Which leads to no specific error counters added, as `set_counters_errors()` relies on `set_oids_errors()`.
We must then call `set_oids_errors()` from `set_counters_errors()`.
I think other interfaces modes are impacted.

garnier-quentin:
set_counters_errors doesn't rely on set_oids_errors.

CPbN:
Ah yes my bad...
This is due to the fact that in `set_counters_errors` I loop through the hash created in `set_oids_errors`...
Sorry for this...

CPbN:
Last question, any thoughts regarding the limitation I talk about in the first post of this PR ?

garnier-quentin:
You can use extend option yes. because we cannot manage all cases.

CPbN:
But how to set a threshold on all options without having to use `--warning-*` and `--critical-*` 31 times ? :)

garnier-quentin:
You cannot.

CPbN:
Will then lead to an outstandingly long command line ^^

garnier-quentin:
You can create a global errors counters in that case (in the plugin). And set the threshold on that.

CPbN:
I tried to add the following function to this Mikrotik interfaces mode :
```
sub check_options {
    my ($self, %options) = @_;

    $self->SUPER::check_options(%options);

    $self->set_oids_errors();

    if (defined($self->{option_results}->{warning_errors})) {
        foreach my $oid (keys %{$self->{oids_errors}}) {
            $oid =~ /^oid_if(In|Out)(.*)$/;
            if (!defined($self->{option_results}->{'warning-' . lc($1) . '-' . lc($2)})) {
                $self->{option_results}->{'warning-' . lc($1) . '-' . lc($2)} = $self->{option_results}->{warning_errors};
            }
        }
    }

    if (defined($self->{option_results}->{critical_errors})) {
        foreach my $oid (keys %{$self->{oids_errors}}) {
            $oid =~ /^oid_if(In|Out)(.*)$/;
            if (!defined($self->{option_results}->{'critical-' . lc($1) . '-' . lc($2)})) {
                $self->{option_results}->{'critical-' . lc($1) . '-' . lc($2)} = $self->{option_results}->{critical_errors};
            }
        }
    }
}
```

I see the various `warning-*` and `critical-*` options set into `$self->{option_results}`.
Unfortunately, they are not applied.
In addition, using `--use-new-perfdata`, they are not rewritten to their new format.
I'm still investigating, but perhaps you have a clue ?

Thx :+1:

garnier-quentin:
I think you need to do it in check_options before the super

CPbN:
I tried to move the SUPER call after my loops, but unfortunately then the various `warning-*` and `critical-*` options are not set anymore into `$self->{option_results}`.
You perhaps meant in `check_options` in `snmp_standard::mode::interfaces` directly ?

CPbN:
OK I managed to correct it thanks to your advice.
I'll push a PR in a few minutes.
Thx :+1:

PR #1516, opened 2019-05-17
Add time mode to Mikrotik

CPbN:
Hi,

This mode simply adds time mode to `network::mikrotik::snmp::plugin`.

Thank you :+1:


garnier-quentin:
I dont understand why you want to add ntp mode ?

CPbN:
So that we can verify that device date/time is correctly set (and is not drifting), comparing to the engine localtime (or to NTP server if provided using `--ntp-hostname`).
I discovered many routers not synced in time, which has some drawbacks such as difficult troubleshooting / log reading, schedule tasks not in time etc...
Monitoring time sync then makes sense :)

CPbN:
Perhaps the `snmp_standard::mode::ntp` should have been better named `snmp_standard::mode::time`, as NTP is just an option to this mode.

garnier-quentin:
Ok i see.

CPbN:
Thank you Quentin :)

PR #1524, opened 2019-05-27
3CX check for updates

CPbN:
Hi,

This PR enhances 3CX plugin checking for internal software updates.

Thank you 👍

### DIFF
--- a/apps/voip/3cx/restapi/custom/api.pm
+++ b/apps/voip/3cx/restapi/custom/api.pm
@@ -194,6 +194,12 @@ sub request_api {
         warning_status => '', unknown_status => '', critical_status => '%{http_code} < 200 or %{http_code} >= 300'
     );
 
+    # Some content may be strangely returned, for example : "[{\"Category\":\"provider\",\"Count\":1}]"
+    if ($content =~ /^"(\[.*\])"$/) {
+        $content = $1;
+        $content =~ s/\\"/"/g;
+    }
+
     my $decoded;
     eval {
         $decoded = JSON::XS->new->utf8->decode($content);
@@ -237,6 +243,20 @@ sub api_system_status {
     my ($self, %options) = @_;
 
     my $status = $self->internal_system_status();
+    return $status;
+}
+
+sub internal_update_checker {
+    my ($self, %options) = @_;
+    
+    my $status = $self->request_api(method => 'GET', url_path =>'/api/UpdateChecker/GetFromParams');
+    return $status;
+}
+
+sub api_update_checker {
+    my ($self, %options) = @_;
+
+    my $status = $self->internal_update_checker();
     return $status;
 }
 
--- a/apps/voip/3cx/restapi/mode/system.pm
+++ b/apps/voip/3cx/restapi/mode/system.pm
@@ -96,8 +96,8 @@ sub new {
     $self->{version} = '1.0';
     $options{options}->add_options(arguments => {
         "unknown-status:s"  => { name => 'unknown_status', default => '' },
-        "warning-status:s"  => { name => 'warning_status', default => '' },
-        "critical-status:s" => { name => 'critical_status', default => '%{health} =~ /false/' },
+        "warning-status:s"  => { name => 'warning_status', default => '%{display} =~ /HasUpdatesAvailable/ && %{health} =~ /true/' },
+        "critical-status:s" => { name => 'critical_status', default => '%{display} !~ /HasUpdatesAvailable/ && %{health} =~ /false/' },
     });
 
     return $self;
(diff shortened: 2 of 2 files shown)


apps/voip/3cx/restapi/custom/api.pm
```diff
@@ -194,6 +194,12 @@ sub request_api {
         warning_status => '', unknown_status => '', critical_status => '%{http_code} < 200 or %{http_code} >= 300'
     );
 
+    # Some content may be strangely returned, for example : "[{\"Category\":\"provider\",\"Count\":1}]"
+    if ($content =~ /^"(\[.*\])"$/) {
+        $content = $1;
+        $content =~ s/\\"/"/g;
+    }
```
garnier-quentin: I can't accept that. It can break json

apps/voip/3cx/restapi/mode/system.pm
```diff
@@ -96,8 +96,8 @@ sub new {
     $self->{version} = '1.0';
     $options{options}->add_options(arguments => {
         "unknown-status:s"  => { name => 'unknown_status', default => '' },
-        "warning-status:s"  => { name => 'warning_status', default => '' },
-        "critical-status:s" => { name => 'critical_status', default => '%{health} =~ /false/' },
+        "warning-status:s"  => { name => 'warning_status', default => '%{display} =~ /HasUpdatesAvailable/ && %{health} =~ /true/' },
+        "critical-status:s" => { name => 'critical_status', default => '%{display} !~ /HasUpdatesAvailable/ && %{health} =~ /false/' },
```
garnier-quentin: Not by default. BUt you can set the threshold in your command line.

CPbN: The previous `critical-status` would then trigger if no update is available, as `HasUpdatesAvailable` would then be false.

Would I then have to rename `HasUpdatesAvailable` to `SystemIsUptodate` ?
An available update would then trigger a critical alert.

apps/voip/3cx/restapi/custom/api.pm
CPbN: The added API call returns `"[{\"Category\":\"provider\",\"Count\":1}]"`.
This is not JSON compliant, this is why we have to remove leading and trailing `"`, and remove escaping `\` characters before `"`, but of course only if `$content` matches this strange content format.

apps/voip/3cx/restapi/mode/system.pm
garnier-quentin: Yes that better. like that, we use the same for 'true' and 'false'.

apps/voip/3cx/restapi/custom/api.pm
garnier-quentin: The JSON is not compliant ? So that's a bug of 3cx ?

CPbN: I think I will not manage to get 3CX to correct this, as it is not documented nor supported :-|
In addition, this is not really a bug, as the JSON string is simply surrounded by `"`, which could in some kind be "expected". We simply have to remove the surrounding...

apps/voip/3cx/restapi/mode/system.pm
CPbN: OK, let me improve this then :) :+1:

CPbN: Should be OK now Quentin :)

apps/voip/3cx/restapi/custom/api.pm
garnier-quentin: Could you provide me the raw string result of that API call ?

CPbN: Here it is :
`"[{\"Category\":\"provider\",\"Count\":1}]"`

CPbN: If no update is available, I get :
`"[]"`

garnier-quentin: The string seems ok. You should can decode with json without problem (no change is needed)

garnier-quentin: You do the print with Data::Dumper ?

CPbN: Without this trick the exception is triggered :
`Cannot decode json response: JSON text must be an object or array (but found number, string, true, false or null, use allow_nonref to allow this)`

CPbN: I print response content using `--debug`.
This is also what curl returns.

CPbN: With Data::Dumper :
`$VAR1 = '"[{\\"Category\\":\\"provider\\",\\"Count\\":1}]"';`

CPbN:
Many thx Quentin 👍

garnier-quentin:
I have changed your code. Could you test it ?

CPbN:
Works nicely, nice trick !
Thx 👍

PR #1530, opened 2019-05-30
Mikrotik, update mode

CPbN:
Hi,

Here's a new mode to Mikrotik plugin to check for RouterOS updates.

Mode can get latest available version by itself.
One can also provide his expected version, for example if he wants his routers to all run a specific version.

Thank you 👍

### DIFF
--- a/network/mikrotik/snmp/mode/update.pm
+++ b/network/mikrotik/snmp/mode/update.pm
@@ -18,7 +18,7 @@
 # limitations under the License.
 #
 
-package network::mikrotik::snmp::mode::version;
+package network::mikrotik::snmp::mode::update;
 
 use base qw(centreon::plugins::mode);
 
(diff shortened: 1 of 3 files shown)


garnier-quentin:
Sorry but for me it's not monitoring directly (more configuration). You can use your mode with dyn-mode option.

CPbN:
Yes it's between monitoring and device configuration...
One may like to have version check in Centreon to be sure all his devices (dozens, hundreds... of devices) are running the expected / last software version. At least I encountered this situation :)
I like to have everything in Centreon Plugins directly :)
But as you mentioned perhaps it's not the best place...

CPbN:
As there are some modes to check for OS pending updates, I wondered :)

garnier-quentin:
You can use centreon-plugins. dyn-mode option is for that.
For OS, it's more "useful"

PR #1537, opened 2019-06-04
Mikrotik, add global error options

CPbN:
Hi,

This PR adds 2 new options to interfaces mode for Mikrotik, `--warning-errors` and `critical-errors`.
They allow to set thresholds for all errors at once.

The PR works, however I'm not sure this is the right way to do, so feel free to comment or modify.

Many thx :+1:

### DIFF
--- a/centreon/plugins/output.pm
+++ b/centreon/plugins/output.pm
@@ -1300,7 +1300,7 @@ Common examples:
 
 =over 4
 
-Sum wrong packets from all interfaces (with interface need  --units-errors=absolute): --extend-perfdata-group=',packets_wrong,sum(packets_(discard|error)_(in|out))'
+Sum bad packets from all interfaces (with interface need  --units-errors=absolute): --extend-perfdata-group=',packets_bad,sum(packets_(discard|error)_(in|out))'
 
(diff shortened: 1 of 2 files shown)


garnier-quentin:
Could you test it ?

CPbN:
Works as expected, many thanks for your review :+1:

CPbN:
One issue though @garnier-quentin.
`output_error_template` is never used.
For example, instead of having :
`Packets In FCSError: 1.23%`
I get :
`Packets In Fcserror: 1.23%`
Whatever I put in `output_error_template`, it does not show in the error output.
Thx :+1:

garnier-quentin:
error_template is not for that.

CPbN:
Even of case of error (warning or critical), `custom_errors_output()` function is triggered to display the error message in the plugin short output.
Instead of using `output_error_template` which should (at least if I understand correctly) be used here.

garnier-quentin:
It s not used for that. Output_error template is for example when you skip the counter

CPbN:
OK, I finally managed to do what I want in #1544. Thank you for your answers.

PR #1545, opened 2019-06-05
Mikrotik, improve error label

CPbN:
Hi,

Here's an alternate method for #1544.
It's better from my point of view as it gives any other interface mode the opportunity to do the same easily.

Thank you 👍


CPbN:
Many thanks again Quentin !

PR #1560, opened 2019-07-02
Correctly display Checkoint swap usage

CPbN:
Hi,

This PR modifies Checkpoint swap usage display, so that physical memory is no more counted here.

Until now, we had, for example for a device with 32GB of swap :
Memory : 128GB
Swap : 160GB

Now we have :
Memory : 128GB
Swap : 32GB

Counters are then the same as returned by `os::linux::snmp::plugin`, memory & swap modes, which also work on these devices.

Thank you 👍


garnier-quentin:
Could you test it ?

CPbN:
Thank you @garnier-quentin :+1:
Just a missing comma at the end of line 145 :)

PR #1571, opened 2019-07-04
Add option to output date in a human readable format

Coolgeek789:
The date was not easily read (format mmddHHMMYY). 
I added an option to make it easier to read (20YY/mm/dd HH:MM)


garnier-quentin:
Thanks for PR. Could you test it ?

PR #1576, opened 2019-07-05
add process checking for local AIX

Coolgeek789:
Add process.pm for local mode on AIX.
It is copied from Linux local mode and small changes were made to adapt to AIX and enhance output.

### DIFF
--- a/os/aix/local/mode/process.pm
+++ b/os/aix/local/mode/process.pm
@@ -1,0 +1,284 @@
+#
+# Copyright 2019 Centreon (http://www.centreon.com/)
+#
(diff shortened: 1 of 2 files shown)


garnier-quentin:
Could you attach a ps command output ?

Coolgeek789:
there you go

[ps.txt](https://github.com/centreon/centreon-plugins/files/3362894/ps.txt)

garnier-quentin:
Could you test it ? I have done some changes. An example option to check the time in seconds
```
--warning-status='%{elapsed} > 1000'
```

PR #1579, opened 2019-07-08
3CX update

CPbN:
Hi,

3CX changed answer to one of the API requests in the last 3CX software update.
This PR then handles this.
While here, I redefined `health` to `error` and `display` to `service`, which seems more understandable.
I also slightly redefined short output, also to a more understandable version.

Thank you 👍


CPbN:
Many thanks @garnier-quentin 👍

PR #1607, opened 2019-08-05
fix thread connected for MariaDB versions

s-duret:


garnier-quentin:
Thanks. But i don't want to do it like that

PR #1615, opened 2019-08-09
Add empty lookup as no lookup

Coolgeek789:
add a case for empty --lookup option (--lookup="") in soap requests.


garnier-quentin:
thanks for the PR. Could you test it ?

Coolgeek789:
seems good to me

garnier-quentin:
Thanks for the feedback

PR #1698, opened 2019-10-23
enh(mssql): pagelifeexpectancy instance counters

pkriko:


Sims24:
As discussed together, I prefer to keep it like this and have a global plan about evolution we might bring one-shot to available mssql checks. 

thanks :)

PR #1699, opened 2019-10-23
Revert "enh(mssql): pagelifeexpectancy instance counters"

Sims24:
Reverts centreon/centreon-plugins#1698


Sims24:
(re)Posting the comment where it is relevant :) 

As discussed together, I prefer to keep it like this and have a global plan about evolution we might bring one-shot to available mssql checks.

thanks :)

Issue #1690, opened 2019-10-22
[InfluxDB] Time calculation


CPbN:
Hi,

Using InfluxDB plugin, I would like to check if the last value of a series is not more than 90s old.

I then tried the following :
```
--plugin=database::influxdb::plugin
--mode=query
--query="count,select count(rta) as count FROM \"centreon\".\"365\".\"host-latency\" where \"host\"='myhost' and time>now()-90s group by host"
--instance='host'
--output='Number of updates in last 90 seconds : %{count}'
--critical-status="%{count} == 0"
```

Unfortunately, this does not work as `count` in InfluxDB returns NULL instead of 0...

My idea is then to select the timestamp of the last inserted value and compare it to current timestamp using `--critical-status` option.

First, I then think we need _epoch_ timestamps to easier achieve this.
We should then need the `&epoch=s` query parameter.

Secondly, I tried to select time field, but I did not manage to do it :
```
--plugin=database::influxdb::plugin
--mode=query
--query="time,select last(rta), time as time FROM \"centreon\".\"365\".\"host-latency\" where \"host\"='myhost' group by host TZ('Europe/Paris')"
--instance='host'
--output='Last timestamp : %{time}'
```
returns `status : skipped (no value(s))`.

Finally, I'm not sure how to make the comparison with current timestamp in `--critical-status` option (`time()` does not seem to be evaluated)...

Any help ?

Thank you very much 👍

garnier-quentin:
Maybe we can use empty value instead of undef.

CPbN:
And I would do `--critical-status="%{count} eq '' || %{count} == 0"` ?

garnier-quentin:
that's the idea

CPbN:
Sounds good, should work :)
Thank you for your quick feedback 👍

garnier-quentin:
Could you test it ?

CPbN:
Unfortunately this does not help.

Here's the raw result (`--debug`) when count returns something >0 :
`{"results":[{"statement_id":0,"series":[{"name":"host-latency","tags":{"host":"myhost"},"columns":["time","count"],"values":[["2019-10-24T12:59:00.609502725Z",1]]}]}]}`
And when it should return 0 but returns null :
`{"results":[{"statement_id":0}]}`

garnier-quentin:
It cannot work. I have looked the code. We don't know the instance. Influxdb provides that information.

garnier-quentin:
[...]

PR #1702, opened 2019-10-24
InfluxDB, allow to select custom field

CPbN:
Hi,

This PR selects the correct InfluxDB data column, given by the `label` field of the query, and not the second (index 1) column (the first one being the time column).
This was first done to be able to select time column, but then allows to select any other one (think about multi column results).
Time column has been moved to epoch time (thanks to the query parameter) for better handling, in thresholds for example.

Refers to issue #1690.

Thank you !


garnier-quentin:
ok for me.

PR #1707, opened 2019-10-30
Add possiblity to use str2time()

CPbN:
Hi,

This PR allows to use `str2time()` in _--convert-custom-values_ option of _string-value_ mode, in _apps::protocols::snmp_ plugin.

For example :
`--convert-custom-values='(time - str2time($value)) / 3600'`

Many thanks 👍

### DIFF
--- a/snmp_standard/mode/stringvalue.pm
+++ b/snmp_standard/mode/stringvalue.pm
@@ -25,7 +25,6 @@ use base qw(centreon::plugins::mode);
 use strict;
 use warnings;
 use centreon::plugins::misc;
-use Date::Parse;
 
 sub new {
     my ($class, %options) = @_;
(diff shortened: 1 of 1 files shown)


garnier-quentin:
I don't want to force at using a specific module. Please add an option to load your module

CPbN:
Thank you @garnier-quentin for your feedback.
Should it be an option specific to this plugin only, or a more global option which would be inherited by all plugins ?

garnier-quentin:
it should be specific

CPbN:
Should be OK with last commit :wink:
We can now do :
`--use-perl-mod='Date::Parse' --convert-custom-values='Date::Parse::str2time($value)'`
And many other things as the option is rather generic.
Thx 👍

CPbN:
Many thx Quentin !

PR #1710, opened 2019-11-05
add cmd-return for AIX (taken from linux)

Coolgeek789:
add cmd-return mode for AIX.
It's just the cmd-return for Linux that I put for AIX


garnier-quentin:
Thanks!

PR #1720, opened 2019-11-12
Display number of partitions

CPbN:
Hi,

This PR re-enables the display of number of partitions monitored in plugin short output.
Useful when regex-selecting FS, so that we can quickly know whether or not the right number of FS is monitored.

Thank you 👍


garnier-quentin:
You have an option for that ;)
--display-ok-counters=count

CPbN:
As simple as that :D
Thx ;)

PR #1727, opened 2019-11-14
Add count to storage mode

CPbN:
Hi,

This PR adds partitions count in `storage` mode.
And corrects a typo (I think) in `diskusage` mode.

Thank you 👍

### DIFF
--- a/snmp_standard/mode/diskusage.pm
+++ b/snmp_standard/mode/diskusage.pm
@@ -49,7 +49,7 @@ sub set_counters {
     ];
     
     $self->{maps_counters}->{global} = [
-        { label => 'count', nlabel => 'storages.partitions.count', display_ok => 0, set => {
+        { label => 'count', nlabel => 'storage.partitions.count', display_ok => 0, set => {
                 key_values => [ { name => 'count' } ],
                 output_template => 'Partitions count : %d',
                 perfdatas => [
--- a/snmp_standard/mode/storage.pm
+++ b/snmp_standard/mode/storage.pm
@@ -162,9 +162,21 @@ sub set_counters {
     my ($self, %options) = @_;
     
     $self->{maps_counters_type} = [
+        { name => 'global', type => 0, cb_init => 'skip_global' },
         { name => 'storage', type => 1, cb_prefix_output => 'prefix_storage_output', message_multiple => 'All storages are ok', skipped_code => { -10 => 1 } },
     ];
     
+    $self->{maps_counters}->{global} = [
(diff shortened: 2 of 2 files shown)


Sims24:
Hi @CPbN 

This modification lead to a more important disk usage (RDD & data_bin).  Could you please share with me the usecase behind this ? When you check several disk path through regexp I guess?

Honnestly I would prefer to have a dynamic counter declaration when it's relevant. Is it a problem for you?

cgagnaire:
The `skip_global` function is missing.

CPbN:
Hi @Sims24 

Sorry for the side effect, I did not think about that...
Having this counter is useful yes using regexp for disk path. You can then quickly know, thanks to the plugin output, if you monitor the expected number of filesystems.

I may remember in the past this sort of counter had to be enabled thanks to an option, does not seem to be the case anymore.

One could also filter this counter out of perfdata using `--filter-perfdata` option.

Note that we also have the same sort of counter in various other plugins, such as `snmp_standard::mode::diskusage`.

Note sure what the best option here, but of course we can make it optional.

Many thanks 👍

CPbN:
> The `skip_global` function is missing.

Quentin removed it himself : https://github.com/centreon/centreon-plugins/commit/6ffa73031febb89ee7751f46c2f1a51fa80d3fb8#diff-7eb447d6b0eb5bd1649ee71cc0956074

CPbN:
And from `storage` mode too actually : https://github.com/centreon/centreon-plugins/commit/a081c921cc712633c2dab694d09729f35c672adb#diff-59a6423dffc93aadac4b1fc10293a88a

Sims24:
Roger, @CPbN  would it be a problem for you if we skip the creation of the count perfdata when only one filesystem/storage match the regex?

CPbN:
Not at all, perfect.
This was the previous behaviour as far as I remember.
Implementing this fix, could you do it on both `storage` and `diskusage` mode please ?
Many thx !

Sims24:
Thanks for your fast feedback @CPbN, yes we will do this on both modes. 

@garnier-quentin could you please reintroduce skip_global at the global counter level and add the missing skip_global function as suggested by @cgagnaire. 

Target behaviour is similar to the CPU here => https://github.com/centreon/centreon-plugins/blob/dccdc34b0a97949526aa1b73174287f04fa19f3c/centreon/common/ingrian/snmp/mode/cpu.pm#L65

Many thxx

garnier-quentin:
I won't do that. Please use --filter-perfdata for that. skip_global is become very shitty with new perfdata.

PR #1754, opened 2019-11-29
Improve Juniper CPU / memory modes

CPbN:
Hi,

This improves Juniper CPU / memory modes.
These 2 were stuck to the Routing Engine.
Whereas we can also check the FPCs / PICs.

Note that `operating` component of `hardware` mode can already check all of them.
But it is not practical at all, this component is too heavy as we can't easily separate what we want to monitor (states, CPU, memory...). For example we can't monitor temperatures only.
IMO it should be split / rewritten, some items such as CPU and memory checks, which are already done here, could perhaps simply be removed, limiting the `operating` component to check state only, as with `alarm` and `fru` components. This would also make it much faster.

Thank you 👍

### DIFF
--- a/network/juniper/common/junos/mode/cpu.pm
+++ b/network/juniper/common/junos/mode/cpu.pm
@@ -67,14 +67,14 @@ sub run {
     my $routing_engine_find = 0;
     my @oids_routing_engine = ();
     foreach my $oid (keys %$result) {        
-        if ($result->{$oid} =~ /$filter/i) {
+        if ($result->{$oid} =~ /$self->{option_results}->{filter}/i) {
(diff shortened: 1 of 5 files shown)


garnier-quentin:
It breaks the compatibility. So i prefer to do it completely (migrate with counters class).

CPbN:
You mean it breaks users using current modes ?

garnier-quentin:
It's done. you can test it

CPbN:
Many thx Quentin, I'll report ASAP 👍

CPbN:
It works perfectly, thank you again !

garnier-quentin:
you're welcome. thanks for the feedback

PR #1759, opened 2019-12-09
Unexpected errors are UNKNOWN

CPbN:
Hi,

You'll tell me, but I think unexpected errors should raise as UNKNOWN, not as CRITICAL.

Thank you 👍


garnier-quentin:
I prefer to use option_exit. We still can set critical.

CPbN:
Yes you're right, nicely done, many thanks Quentin :+1:

PR #1789, opened 2020-01-06
Add missing option to set SSL options

alfredosilvestre-natixis:
When using this plugin I noticed that since our prime had a self-signed certificate, the plugin was not taking this into account and this fixes the issue.

### DIFF
--- a/network/cisco/prime/restapi/custom/api.pm
+++ b/network/cisco/prime/restapi/custom/api.pm
@@ -48,7 +48,7 @@ sub new {
             "username:s@" => { name => 'username' },
             "password:s@" => { name => 'password' },
             "timeout:s@"  => { name => 'timeout' },
-			"ssl-opt:s@"  => { name => 'ssl_opt' },
+            "ssl-opt:s@"  => { name => 'ssl_opt' },
         });
     }
     $options{options}->add_help(package => __PACKAGE__, sections => 'REST API OPTIONS', once => 1);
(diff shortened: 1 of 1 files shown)


garnier-quentin:
You can already used that option:
```
Backend lwp Options:
    --ssl-opt
            Set SSL Options (--ssl-opt="SSL_version => TLSv1"
            --ssl-opt="SSL_verify_mode => SSL_VERIFY_NONE").

    --ssl   Set SSL version (--ssl=TLSv1).

Backend Curl Options:
    --curl-opt
            Set CURL Options (--curl-opt="CURLOPT_SSL_VERIFYPEER => 0"
            --curl-opt="CURLOPT_SSLVERSION => CURL_SSLVERSION_TLSv1_1" ).
```

alfredosilvestre-natixis:
Yes, but it is not working with this specific plugin "network::cisco::prime::restapi::custom::api"

Output is -> Unknown option: ssl-opt

garnier-quentin:
For me, it's working:
```
perl centreon_plugins.pl --plugin=network/cisco/prime/restapi/plugin.pm --mode=ap-usage --ssl-opt="tt" --hostname="test" 
```

Which version do you use ? If you use --http-backend=curl, it's --curl-opt

alfredosilvestre-natixis:
We had an old version of the http plugin, after updating everything works.

Thanks

PR #1796, opened 2020-01-10
Update memory.pm

su-rdi:
sub compat_threshold_counter was missing


su-rdi:
older rev result unknow option when using --warning-usage or --critical-usage
Unknown option: warning-usage at /usr/lib/centreon/plugins/centreon/plugins/alternative/Getopt.pm line 67.

garnier-quentin:
It's not missing. It's in: centreon::plugins::templates::counter
I think you don't use the last version

PR #1800, opened 2020-01-14
Add Dynatrace plugin

CPbN:
Hi,

Here is a [Dynatrace](https://www.dynatrace.com) plugin.
It checks for problems ([API ref](https://www.dynatrace.com/support/help/extend-dynatrace/dynatrace-api/environment-api/problems/problems/get-feed/)) thanks to the counter template.

Thank you 👍

### DIFF
--- a/apps/dynatrace/restapi/mode/problem.pm
+++ b/apps/dynatrace/restapi/mode/problem.pm
@@ -53,7 +53,20 @@ sub set_counters {
     my ($self, %options) = @_;
 
     $self->{maps_counters_type} = [
+        { name => 'global', type => 0 },
         { name => 'problem', type => 1, cb_prefix_output => 'prefix_service_output', message_multiple => 'No problem' }
+    ];
+
+    $self->{maps_counters}->{global} = [
+        { label => 'problems-open', nlabel => 'problems.open.current', set => {
+                key_values => [ { name => 'problems_open' } ],
+                output_template => 'problems open : %s',
+                perfdatas => [
+                    { label => 'problems_open',  template => '%s', value => 'problems_open_absolute',
+                      min => 0 },
+                ],
+            }
+        },
     ];
 
     $self->{maps_counters}->{problem} = [
@@ -103,6 +116,7 @@ sub manage_selection {
     my ($self, %options) = @_;
 
     my $problem = $options{custom}->api_problem(relative_time => $options{options}->{relative_time});
+    my $problems_open = 0;
     $self->{problem} = {};
     foreach my $item (@{$problem}) {
         $self->{problem}->{$item->{displayName}} = {
@@ -117,7 +131,14 @@ sub manage_selection {
             endTime => $item->{endTime} > -1 ? $item->{endTime} / 1000 : -1,
             commentCount => $item->{commentCount},
         };
+        if ($item->{status} eq 'OPEN') {
+            $problems_open++;
+        }
     }
+
+    $self->{global} = {
+        problems_open => $problems_open,
+    };
 }
 
 1;
@@ -150,6 +171,11 @@ Can use special variables like: %{status}, %{impactLevel}, %{severityLevel}, %{e
 
 Set critical threshold for status (Default: '%{status} eq "OPEN"').
 Can use special variables like: %{status}, %{impactLevel}, %{severityLevel}, %{entityName}, %{eventType}, %{entityId}, %{startTime}, %{endTime}, %{commentCount}
+
+=item B<--warning-*> B<--critical-*>
+
+Thresholds.
+Can be: 'problems-open'.
 
 =back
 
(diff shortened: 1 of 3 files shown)


Sims24:
Hi @CPbN 

Thanks for all your contributions :) That's simply great ! Quentin is prelly busy these days but we will discuss your PRs by the end of the month! 

We would love to help you to extend this one with other modes :) Looks like there is a lot of interesting things in their environment endpoints (https://www.dynatrace.com/support/help/extend-dynatrace/dynatrace-api/).

CPbN:
Hi @Sims24,

Many thanks for your feedback 👍
Glad to hear you'll take a look at these PRs !
And happy to contribute 😃

You're right regarding Dynatrace API, there are other endpoints / modes which may be nice to have.
For now I've then have implemented the most interesting / critical one, the `problem` mode.
I may dig into other mode(s), if so I'll push into this PR :+1:

Thank you again !

apps/dynatrace/restapi/mode/problem.pm
```diff
@@ -0,0 +1,160 @@
...
+            eventType => join(",", uniq map { "$_->{eventType}" } @{$item->{rankedImpacts}}),
+            entityId => join(",", uniq map { "$_->{entityId}" } @{$item->{rankedImpacts}}),
+            startTime => $item->{startTime} / 1000,
+            endTime => $item->{endTime} > -1 ? $item->{endTime} / 1000 : -1,
+            commentCount => $item->{commentCount},
+        };
+    }
+
+use Data::Dumper;
+print Dumper($self->{problem});
```
tanguyvda: 😛

CPbN: Thank you @tanguyvda for this 👍

apps/dynatrace/restapi/mode/problem.pm
```diff
@@ -0,0 +1,156 @@
...
+
+use strict;
+use warnings;
+use List::MoreUtils qw(uniq);
+use centreon::plugins::templates::catalog_functions qw(catalog_status_threshold);
+
+sub custom_status_output { 
+    my ($self, %options) = @_;
+
+    my $msg = '';
```
garnier-quentin: I think, a message is missing

CPbN: Well, this produces output like this :
`WARNING: Problem '382'`
`CRITICAL: Problem '267'`
I deliberately did not put any additional info after the problem number.

CPbN:
> Could you add a counter with the number of open issues ?

Sure, good idea 👍

apps/dynatrace/restapi/mode/problem.pm
garnier-quentin: Please use 'type => 2' counters

CPbN:
> Could you add a counter with the number of open issues ?

Just added it, however this will be the global number of open problems, not the one matching the rules given by the user through the `--*-status` options.

apps/dynatrace/restapi/mode/problem.pm
CPbN: With type 2 I'm afraid we will loose problem number in short output ?
It is rather convenient, as we can use it in Dynatrace to search for the concerned problem.

CPbN: I then tried to improve short output, of course feel free if needed.

CPbN: The only drawback is that when there is only on problem opened which does not trigger the `WARNING` or `CRITICAL` condition, then output is like this :
```
centreon_plugins.pl --plugin=apps::dynatrace::restapi::plugin --mode=problem --api-password=$pw --critical-status='(time-%{startTime} > 3600) && %{status} eq "OPEN" && %{commentCount} == 0' --environment-id=cpbn
OK: Number of open problems : 1 - Problem '831' detected | 'problems.open.count'=1;;;0;
```
Which does not sound so much logical.
Could still play with `--change-short-output` to replace with the `message_multiple` (_No problem detected_), but perhaps there's a better solution.
Expected output would have then been :
`OK: Number of open problems : 1 - No problem detected | 'problems.open.count'=1;;;0;`

CPbN:
Many thx Quentin 👍

garnier-quentin:
I have done some changes. Could you test it ?

CPbN:
It works perfectly, thank you !
The only drawback now that you moved to type 2 is that we do not have the problem number anymore in the short output :
`CRITICAL: 1 problem(s) detected | 'problems.open.count'=1;;;0; 'problems'=1;;;0;`
Which was convenient, as we can quickly find the problems thanks to its number in the Dynatrace interface. Any way to have it in the output ? Do you think this would be relevant enough ?
Thank you 👍

garnier-quentin:
it's in the verbose output

garnier-quentin:
if you have X problems, it will be too much for the short output

PR #1809, opened 2020-01-18
failoverstatus.pm

johnhighway:
Dropping additional code.


garnier-quentin:
Could you test the mode ?

PR #1812, opened 2020-01-20
Change memory calculation for redhat 7.1

mullergg:
According to : https://access.redhat.com/solutions/406773

I added a function to find what redhat version we have and then adapt the calculation method.

### DIFF
--- a/os/linux/local/mode/memory.pm
+++ b/os/linux/local/mode/memory.pm
@@ -1,5 +1,5 @@
 #
-# Copyright 2020 Centreon (http://www.centreon.com/)
+# Copyright 2019 Centreon (http://www.centreon.com/)
 #
 # Centreon is a full-fledged industry-strength solution that meets
 # the needs in IT infrastructure and application monitoring for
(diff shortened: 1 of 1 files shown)


garnier-quentin:
Why do you want to change it for rhel 7.1 ?

garnier-quentin:
Ok. I'll take a look soon (thursday)

mullergg:
Thanks

garnier-quentin:
Could you test it ?
For me, i think we should use SReclaimable no ?

mullergg:
The check of the red hat version is wrong.
The regex does not match on Redhat because it's Red Hat in the file.
It is also wrong because you don't match on version like 8.0 or every version > 7.1 cotaining a 0 in minor.

Do you also remove the Slab value in the output and perfdata on purpose ?

I also hesitate to do the calculation on SReclaimable value, because the command free is based on this value. But Red Hat seems to say that the real RAM consumption is based on Slab. 
So I don't know which one has the right answer.

garnier-quentin:
Sorry for the delay. i have fixed it.

mullergg:
Ok thanks, I will try to test it soon.

mullergg:
Hello !
It seems to works, tanks for the update.
Did you analyze if we have to use Slab or SReclaimable ?

garnier-quentin:
we are going to use what rhel tell us in the doc.

PR #1828, opened 2020-02-03
Allow newline character in parameters

UrBnW:
Hi,

This PR allows to use parameters with embedded newlines.
This works when using Perl `Getopt` (`$alternative = 0` in `centreon/plugins/options.pm`).
But it was not working with `centreon::plugins::alternative::Getopt`.

This is now fixed thanks to this PR 👍

Thx !


UrBnW:
Of course this is related to #1827 :)

PR #1839, opened 2020-02-11
snmp plugin for listing process with their memory and cpu stats

xdrive05:
Memory and CPU % stats for processes for Linux hosts


garnier-quentin:
I added --add-stats option:
perl centreon_plugins.pl --plugin=os/linux/snmp/plugin.pm --mode=list-processes --hostname=127.0.0.1 --add-stats

And to get perfdata, you should use --mode=processcount

garnier-quentin:
If you need something else, tell me

xdrive05:
Wow, thanks for the fast reply.

Pretty close to what I'm attempting to do here and actually that change is almost identical to one of the first passes I made at this before I realized we were going to need output in nagios perfdata format for ingestion.

Our use case is probably not all that common, but this came from a request from our users to be able to see cpu % and memory usage from each discrete processes/thread running on a host.

My attempt is definitely a dirty hack but the idea was to combine functionality from both list-processes and processcount into one.

Feeding process names from list-processes into processcount to get the perfdata almost got us what we needed, but it was a bit slow querying everything individually. 
Also because processcount --process-name='name' appears to aggregate everything by process name, we weren't able to get the breakdown we needed from multiple process with the same name or any child processes.

Thanks again!

garnier-quentin:
You want cpu and memory without aggregate ?

xdrive05:
Pretty much. Cpu % and memory usage for each pid basically.

My example is crude, but gets us output like this for ingestion: 
`OK: List processes stats: | '2340_kworker/3:1H_name'=kworker/3:1H;;;; '2340_kworker/3:1H_path'=;;;; '2340_kworker/3:1H_parameters'=;;;; '2340_kworker/3:1H_type'=operatingSystem;;;; '2340_kworker/3:1H_pid'=2340;;;; '2340_kworker/3:1H_status'=runnable;;;; '2340_kworker/3:1H_cpu'=0.00 %;;;; '2340_kworker/3:1H_mem'=0;;;; '30718_sshd_name'=sshd;;;; '30718_sshd_path'=sshd: root@pts/0;;;; '30718_sshd_parameters'=;;;; '30718_sshd_type'=application;;;; '30718_sshd_pid'=30718;;;; '30718_sshd_status'=runnable;;;; '30718_sshd_cpu'=0.00 %;;;; '30718_sshd_mem'=5668864;;;; '32_kintegrityd_name'=kintegrityd;;;; '32_kintegrityd_path'=;;;; '32_kintegrityd_parameters'=;;;; '32_kintegrityd_type'=operatingSystem;;;; '32_kintegrityd_pid'=32;;;; '32_kintegrityd_status'=runnable;;;; '32_kintegrityd_cpu'=0.00 %;;;; '32_kintegrityd_mem'=0;;;; '5071_snmpd_name'=snmpd;;;; '5071_snmpd_path'=/usr/sbin/snmpd;;;; '5071_snmpd_parameters'=-LS0-6d -f;;;; '5071_snmpd_type'=application;;;; '5071_snmpd_pid'=5071;;;; '5071_snmpd_status'=running;;;; '5071_snmpd_cpu'=0.04 %;;;; '5071_snmpd_mem'=6197248;;;;`

garnier-quentin:
need to think about it. but i can't merge in that state.

xdrive05:
Understandable.
I thought this might be useful if anyone else is looking for something similar but I also know it's pretty rough code.
I probably should have made this a feature request instead of a pull request.
Thanks for taking a look!

Issue #1846, opened 2020-02-13
Uptime of 0 leads to empty output


UrBnW:
Hi,

Uptime mode leads to following output :
`WARNING: System uptime is:  | 'uptime'=0s;1800:;;0;`

When :
```
#  snmpwalk (...) .1.3.6.1.2.1.25.1.1.0
.1.3.6.1.2.1.25.1.1.0 = Timeticks: (0) 0:00:00.00
```

I face issues on some devices with SNMP hanging and returning this, this is why I had time to investigate this.

I wonder `centreon::plugins::misc::change_seconds` could be the culprit ?
I verified, `$value` is not empty, it's equal to `0`.
https://github.com/centreon/centreon-plugins/blob/9bed462fdd6dd238f1cbad094b2745c03cfc1f37/snmp_standard/mode/uptime.pm#L140

Thx 👍

garnier-quentin:
It's fixed!

UrBnW:
And it works, many thx 👍

UrBnW:
Do you think it could be relevant to return `UNKNOWN` when uptime is 0 ?
Would help to have this return code on such devices where SNMP agent hangs.
Thx !

garnier-quentin:
Maybe a warning you can raise.

UrBnW:
Mmmh I already raise WARNING for values < 1800 seconds.
In terms of threshold options we have :
```
    --warning
            Threshold warning in seconds.
    --critical
            Threshold critical in seconds.
```
I'm not sure a `--unknown` option would make sense.

UrBnW:
Same issue when `$disant_time` returns 0 in time (ntp) mode on such devices :
https://github.com/centreon/centreon-plugins/blob/dc74f419888a2bddf67ca5572a81942bb05e6a94/snmp_standard/mode/ntp.pm#L116

UrBnW:
Same issue when `$disant_time` returns 0 in time (ntp) mode on such devices :
https://github.com/centreon/centreon-plugins/blob/dc74f419888a2bddf67ca5572a81942bb05e6a94/snmp_standard/mode/ntp.pm#L116

It then produces such output :
`CRITICAL: Time offset -1581603667 second(s): Local Time : 1970-01-01T00:00:00 (+0000)`

`UNKNOWN` may perhaps be more relevant here.

garnier-quentin:
you can't have unknown for threshold metrics.

PR #1850, opened 2020-02-13
Add an UNKNOWN option for 0 timestamps

UrBnW:
Hi,

This PR is a proposal which follows our conversation there : https://github.com/centreon/centreon-plugins/issues/1846

It adds an `--unknown-if-zero` option with allows `uptime` and `ntp` nodes to return `UNKNOWN` if time returned by the SNMP agent is 0.

Thx 👍

### DIFF
--- a/snmp_standard/mode/ntp.pm
+++ b/snmp_standard/mode/ntp.pm
@@ -62,9 +62,10 @@ sub new {
     bless $self, $class;
     
     $options{options}->add_options(arguments => { 
-        'ntp-hostname:s' => { name => 'ntp_hostname' },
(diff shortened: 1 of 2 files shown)


garnier-quentin:
Maybe we can add a unknown threshold. But not like that

UrBnW:
OK, thank you for your feedback Quentin.
Let's then find a proper way to do this :)

UrBnW:
_Note that, for the `ntp` mode, the `UNKNOWN` return code was not on the final offset value, but on the intermediate server time, which can easily be compared to 0._

PR #1853, opened 2020-02-17
Add trunk members to the output of each interface

alfredosilvestre-natixis:
Hello,

@fernandoeng has managed a solution to retrieve the trunk members and display them in the output for each interface.

If you find this useful, you can integrate it in the plugin.

Thanks

### DIFF
--- a/network/f5/bigip/snmp/mode/listtrunks.pm
+++ b/network/f5/bigip/snmp/mode/listtrunks.pm
@@ -59,8 +59,43 @@ my $sysTrunkName = '.1.3.6.1.4.1.3375.2.1.2.12.1.2.1.1';
 my $sysTrunkStatus = '.1.3.6.1.4.1.3375.2.1.2.12.1.2.1.2';
 my $sysTrunkOperBw = '.1.3.6.1.4.1.3375.2.1.2.12.1.2.1.5';
 
+my $sysTrunkCfgMember      = '.1.3.6.1.4.1.3375.2.1.2.12.3';
+my $sysTrunkCfgMemberTable = '.1.3.6.1.4.1.3375.2.1.2.12.3.2';
+my $sysTrunkCfgMemberName  = '.1.3.6.1.4.1.3375.2.1.2.12.3.2.1.2';
+
+my $ifDescr = '.1.3.6.1.2.1.2.2.1.2';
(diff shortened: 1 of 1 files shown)


garnier-quentin:
Sorry i don't really see the link with the mib file. Could you provide a full snmpwalk of your F5 ? (to my email):
snmpwalk -ObentU -v 2c -c community IP_F5 .1 > f5_bigip.snmpwalk

alfredosilvestre-natixis:
I believe that you might only need the relevant parts of the MIB (attached) as the full MIB is 14mb

[f5_bigip-IF-MIB.txt](https://github.com/centreon/centreon-plugins/files/4623440/f5_bigip-IF-MIB.txt)

[f5_bigip-sysTrunks.txt](https://github.com/centreon/centreon-plugins/files/4623439/f5_bigip-sysTrunks.txt)

The plugin will add the members in the extended status information like in the example below:
'TRUNK_DMZ_PROD_n_QUALIF' [status: up] [speed: 20000] [members: 5.0(up) 6.0(up)]

garnier-quentin:
Ok i see. But what do you want to monitor exactly ? You have a global status for a F-5 trunk.
You want to monitor status of each interfaces of a trunk ?
Maybe i can update the mode=trunks and add option --add-interface-status:
- monitor status of each interfaces
- number of interfaces found for a trunk (so if you have 0, you can add an alert).

Seems nice for you ?

alfredosilvestre-natixis:
Yes, the idea is to have extra information on the status of the trunk member interfaces as you pointed out.

Since you already have the information for the status and speed of the trunk, we added the members and their status as well -> [members: 5.0(up) 6.0(up)]

The solution in the PR is working on our environment, the option you suggested seems good to let users choose if they want to display this information 😄

garnier-quentin:
It's working but list-trunks is not done for that.

alfredosilvestre-natixis:
Since the mode=trunks has performance data on the trunks, the list-trunks would just have extra information on the trunk members. The option can also be added to both modes as well.

But as you wish, it was just a suggestion, the idea is to have the trunk member information somewhere because as it is there is no check if one member is down.

garnier-quentin:
You can test it with option ```--add-interfaces```

PR #1856, opened 2020-02-18
Uptime, add unknown threshold

UrBnW:
Hi,

This PR is a proposal which follows our conversation there #1846 and there #1850.

It adds an `--unknown` threshold to `uptime` mode.

Thx 👍


UrBnW:
This is to be able to avoid this situation (which I just faced again) :
`WARNING: System uptime is: 0d`
And be able to rather return :
`UNKNOWN: System uptime is: 0d`

PR #1857, opened 2020-02-18
NTP, return unknown in case of 0 timestamp

UrBnW:
Hi,

This PR is a proposal which follows our conversation there #1846 and there #1850.

It modifies the `ntp` mode so that it will return `UNKNOWN` in case remote timestamp is 0 (which should not happen, a host can't be stuck in 1970).

Thx 👍


UrBnW:
This is to be able to avoid this situation (which I just faced again) :
`CRITICAL: Time offset -1591211806 second(s): Local Time : 1970-01-01T00:00:00 (+0000)`
And then to rather return :
`UNKNOWN: Couldn't get system date: Local Time: 0`

PR #1886, opened 2020-03-17
Azure Logs analytics

Tpo76:
I didn't find a way to overload the header "content-type" between authentication and the query request in the api.pm file correctly.
So this one is more a workaround to make this check works. If you know how to do it, I'll make the change.

Possible enhancement: matching string for threshold.

### DIFF
--- a/cloud/azure/custom/api.pm
+++ b/cloud/azure/custom/api.pm
@@ -153,7 +153,7 @@ sub settings {
 
     $self->build_options_for_httplib();
     $self->{http}->add_header(key => 'Accept', value => 'application/json');
-    $self->{http}->add_header(key => 'Content-Type', value => 'application/x-www-form-urlencoded');
+#    $self->{http}->add_header(key => 'Content-Type', value => 'application/x-www-form-urlencoded');
     if (defined($self->{access_token})) {
         $self->{http}->add_header(key => 'Authorization', value => 'Bearer ' . $self->{access_token});
     }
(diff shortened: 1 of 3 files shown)


Sims24:
Hi, 

Thanks, could you provide by mail curl commands from authentication to the query toward .io log endpoint please ? 

Think it could be great to check what custom query we may offer through the cli params.

cloud/azure/custom/api.pm
```diff
@@ -163,7 +163,7 @@ sub settings {
 sub get_access_token {
     my ($self, %options) = @_;
 
-    my $has_cache_file = $options{statefile}->read(statefile => 'azure_api_' . md5_hex($self->{subscription}) . '_' . md5_hex($self->{tenant}) . '_' . md5_hex($self->{client_id}));
+    my $has_cache_file = $options{statefile}->read(statefile => 'azure_api_' . md5_hex($self->{subscription}) . '_' . md5_hex($self->{tenant}) . '_' . md5_hex($self->{client_id}) . '_' . md5_hex($self->{management_endpoint}));
```
garnier-quentin, line 166: Why do you add $self->{management_endpoint} ?

garnier-quentin:
I have fixed the issue with the content-type. Like Sims says, we need to the debug of your endpoint log please.

garnier-quentin:
I have done the change. Could you test with that version please ?

PR #1898, opened 2020-03-24
fix: remove colon char from iface name in packet-errors for Linux local

omercier:


garnier-quentin:
Could you provide the ip addr output with the issue ? Because it should work.

omercier:
HI,
I cannot give you an IP address you will be able to reach, but you can reproduce on any CentOS7 host with a clone of centreon-plugins:
`/usr/lib/centreon/plugins/centreon-plugins/centreon_plugins.pl --plugin os::linux::local::plugin --mode packet-errors --name=".*" --regexp --warning-in-error="" --critical-in-error="" --warning-out-error="" --critical-out-error="" --warning-in-discard="" --critical-in-discard="" --warning-out-discard="" --critical-out-discard="" --filter-state=""  --statefile-dir=/var/log/nrpe/centplugins --no-loopback --verbose`

Replace `--statefile-dir=/var/log/nrpe/centplugins` with the appropriate value, of course.

PR #1908, opened 2020-03-28
fix(disk/storage): output size skipped disks only in debug mode

UrBnW:
Hi,

Some systems may return a lot of 0 byte size disks.
It can then be hard to filter all them out.
Which then messes-up the verbose output.

This PR then solves this.

Thank you 👍


garnier-quentin:
Could you test it ?

UrBnW:
Your fix works perfectly, many thanks 👍

PR #1909, opened 2020-03-28
MikroTik, add proper disk mode

UrBnW:
Hi,

This PR adds a proper MikroTik disk mode.
Until now we could use the memory mode to achieve this, but then we had to change the "RAM" prefix into "Disk" in the short output, and this is not possible in the long output.

As it's almost a copy of the `memory` mode, I hesitated to rename `memory` mode to `storage` mode, and to add a sort of `--type` option, which would allow to tune the prefix, `RAM`, `Disk`, or whatever...

Feel free to tell me what's the best method.

Thank you 👍

### DIFF
--- a/network/mikrotik/snmp/mode/disk.pm
+++ b/network/mikrotik/snmp/mode/disk.pm
@@ -18,7 +18,7 @@
 # limitations under the License.
 #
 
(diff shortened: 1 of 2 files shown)


UrBnW:
Thank you @garnier-quentin 😉

PR #1915, opened 2020-04-01
fix(mem): better handle RedHat specific case

UrBnW:
Hi,

Some RedHat (or CentOS) servers may already have Net-SNMP with the new behavior, but still running an "old" kernel...
`--autodetect-redhat` then does not detect it.
Here's then an additional trick to automatically detect this new behavior.

Thank you 👍


garnier-quentin:
Could you test it ? It should not break.

UrBnW:
Works fine, thank you 👍

PR #1917, opened 2020-04-01
enh(disk): support 32 bits counters

UrBnW:
Hi,

Some (old ?) systems do not provide `High` and `Low` disk usage counters.
Let's then detect if 64 bits counters are available, and use 32 bits ones if not.

Thank you 👍


garnier-quentin:
Could you test on your old systems ?

UrBnW:
Works perfectly too 👍
Thank you again for your reviews !

PR #1920, opened 2020-04-02
add support for ACS8000 series (psu oid differs from ACS6000 series)

toxiczka:
This is mostly a copy of the ACS6000 module with few changes, to support ACS8000 series. Changes are of course the naming (6000 > 8000) and the new OIDs of the power supplys. Tested with all modes against an ACS8000 of course.

### DIFF
--- a/hardware/kvm/avocent/acs/8000/snmp/mode/components/psu.pm
+++ b/hardware/kvm/avocent/acs/8000/snmp/mode/components/psu.pm
@@ -1,0 +1,71 @@
+#
+# Copyright 2019 Centreon (http://www.centreon.com/)
+#
+# Centreon is a full-fledged industry-strength solution that meets
+# the needs in IT infrastructure and application monitoring for
+# service performance.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


garnier-quentin:
Could you provide mib files and full snmpwalk ? I will try to put in the same maybe.

toxiczka:
Here you are! MIB Files are directly downloaded from vertiv.com.

[ACS8000-MIB.asn.txt](https://github.com/centreon/centreon-plugins/files/4421681/ACS8000-MIB.asn.txt)
[snmpwalk.log](https://github.com/centreon/centreon-plugins/files/4421714/snmpwalk.log)

garnier-quentin:
Datas are missing in your snmpwalk. Need following branch:
.1.3.6.1.4.1.10418

toxiczka:
Sorry, my mistake. I've updated the file in my previous comment with the branch you requested.

garnier-quentin:
maybe we can add to check serial ports: status, traffic in/out ?

toxiczka:
Would be great, but far above my skill level.

garnier-quentin:
You can test the new mode

toxiczka:
Have tested today. The new mode works - at least it shows the ports and 0 traffic. Since there seems to be no current activity on this device and I've no other than SNMP access to it, I cannot test if the traffic counters really work.

garnier-quentin:
Thanks for the feedback. it should ;)

PR #1922, opened 2020-04-03
enh(nsclient): default to SSL_VERIFY_NONE

UrBnW:
Hi,

This defaults NSClient++ API to `SSL_VERIFY_NONE`.
Who replace their self-signed certificate by a CA-signed one ? ^^

Thx 👍


garnier-quentin:
For the security, it's not recommended.

UrBnW:
There are some modes where you force it, so I though this was relevant.

garnier-quentin:
Very few (maybe ilo. don't think we can change it).

PR #1925, opened 2020-04-05
new(plugin) add sfdc minimal check

Sims24:


Sims24:
Sample usage / output: 
```

perl centreon_plugins.pl --plugin=apps::salesforce::restapi::plugin --mode=instance-status --hostname=api.status.salesforce.com --http-backend=curl --instance=mysfdcinstance --verbose --alias
OK: Salesforce 'mysfdc' instance status is 'OK' (active:'true') , 0 incidents currently | 'mysfdc#salesforce.incident.current.count'=0;;;0;
Salesforce 'mysfdc' instance status is 'OK' (active:'true') , 0 incidents currently
```

PR #1936, opened 2020-04-13
Rausers mode checkpoint

pkriko:
New mode for checkpoint.

`[root@central centreon-plugins]# perl centreon_plugins.pl --plugin network::checkpoint::snmp::plugin --mode ra-users --hostname=127.0.0.1 --snmp-port=1616 --snmp-community=checkpoint --snmp-autoreduce --buggy-snmp
CRITICAL: Remote User 'pkriko' status : down | 'ra.users.total.count'=510;;;0;
`
`[root@central centreon-plugins]# perl centreon_plugins.pl --plugin network::checkpoint::snmp::plugin --mode ra-users --hostname=127.0.0.1 --snmp-port=1616 --snmp-community=checkpoint --snmp-autoreduce --buggy-snmp
OK: All Remote Remote Access users tunnel are OK | 'ra.users.total.count'=510;;;0;
`
Didn't find the OID .1.3.6.1.4.1.2620.500.9000.1.2 in the MIB but seems to be the username.

### DIFF
--- a/network/checkpoint/snmp/mode/rausers.pm
+++ b/network/checkpoint/snmp/mode/rausers.pm
@@ -53,7 +53,7 @@ sub set_counters {
 
     $self->{maps_counters}->{ratunnel} = [
         { label => 'status', threshold => 0, set => {
-                key_values => [ { name => 'type' }, { name => 'status' }, { name => 'display' } ],
+                key_values => [ { name => 'display' }, { name => 'status' } ],
                 closure_custom_calc => \&catalog_status_calc,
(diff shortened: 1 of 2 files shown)


garnier-quentin:
Do you think it's important to know the status of tunnel users ? User entries are removed when it disconnects ?

Sims24:
@garnier-quentin yes they are removed. Therefore I'm not sure the status is relevant, even if catching a down tunnel might be interesting (When it can happen, can't say maybe a transition state before table content cleaning?)

garnier-quentin:
Could you test ?

PR #1942, opened 2020-04-16
Better handle Blade/Flex temp thresholds

UrBnW:
Hi,

Strange case, but just faced it.

Thx 👍

### DIFF
--- a/centreon/common/ibm/nos/snmp/mode/components/temperature.pm
+++ b/centreon/common/ibm/nos/snmp/mode/components/temperature.pm
@@ -47,6 +47,11 @@ sub check {
     $temperature = $1 if ($1 > $temperature);
     my $critical_mib = ($warning_mib + $2) / 2;
 
+    if ($warning_mib == $critical_mib) { #seen on some chassis !
+        $warning_mib -= 10;
+        $critical_mib -= 5;
+    }
+
     $self->{components}->{temperature}->{total}++;
 
     $self->{output}->output_add(long_msg => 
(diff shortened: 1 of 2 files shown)


garnier-quentin:
Why we don't use the shutdown value and that's all ?

UrBnW:
Because the device will shutdown, and it will be too late :)

garnier-quentin:
That formula is weird: my $critical_mib = ($warning_mib + $2) / 2;
Why not:
$critical_mib = $2 - 5;
if ($critical_mib < $warning_mib) { $critical_mib = $warning_mib }

UrBnW:
On the various devices I have, generally, `shutdown` temperature is `warning` temperature + 10.
I thought it was a good rule to have `CRITICAL` threshold set by default between these 2 values, so that you still have some time before the device will shutdown.

I faced however some devices where `shutdown` temperature == `warning` temperature.
Configuration error ? Perhaps, I don't know.
I then applied the -10 / -5 as generally seen on other devices.

If you're not please with `/ 2`, I would suggest :
```
$result->{hwInfoTemperatureWarn} =~ /^([.0-9]+)\s*C\s*\(Warn(?:ing)?\s*at\s*([.0-9]+)\s*C/i;
my $temperature = $1;
$result->{hwInfoTemperatureShut} =~ /^([.0-9]+)\s*C\s*\(Shutdown\s*at\s*([.0-9]+)\s*C/i;
$temperature = $1 if ($1 > $temperature);
my $critical_mib = $2 - 5;
my $warning_mib = $critical_mib - 5;
```

UrBnW:
Let me know if you ant to go this way, I can quickly update this PR 👍

garnier-quentin:
ok i understand

UrBnW:
Thanks again for your reviews 😉

PR #1962, opened 2020-04-23
Disks, keep small SNMP requests

UrBnW:
Hi,

This PR tries to keep `disk-usage` mode SNMP requests as small as possible, adding a `--force-counters32` option to grab only needed OIDs, to improve performance.

Thx 👍


UrBnW:
I went for a `--force-counters32` option, should be "OK" then :)

PR #1985, opened 2020-05-05
Interfaces, add --force-counters64 option

UrBnW:
Hi,

Here's a PR to add a new `--force-counters64` option to interfaces mode.
This will then force mode to grab 64bits counters only (when both 32 and 64 bits are available).
This for performance reasons.

As an example, on a device where 5184 OIDs are retrieved, it takes 32 seconds.
With `--force-counters64`, "only" 3240 OIDs are retrieved, and it takes 19 seconds.
Nice improvement then, as in #1962.

Thank you 👍


garnier-quentin:
Could you test ?

UrBnW:
Sure, let me give this a test 👍

UrBnW:
Works as expected as per my tests, perfect, many thanks !

PR #1996, opened 2020-05-13
Bugfix on infinite loop

alfredosilvestre-natixis:
When testing with a ciscoprime with more than 1000 APs, the function stays in an infinite loop.
This is blocked by the setting "nbi.rateLimit.perUserThreshold" and issues the error:
"NBI Rate limit for user exceeded (More than 5 in 1,000 ms)"


garnier-quentin:
thanks for the patch

PR #1999, opened 2020-05-14
fix(output) set prct to 0 if max is 0

UrBnW:
Hi,

A simple fix to get a correct `0%` metric value when max value is `0`.

Thx 👍

### DIFF
--- a/centreon/plugins/output.pm
+++ b/centreon/plugins/output.pm
@@ -1019,10 +1019,14 @@ sub apply_pfdata_invert {
 sub apply_pfdata_percent {
     my ($self, %options) = @_;
 
-    return if (!defined(${$options{perf}}->{max}) || ${$options{perf}}->{max} eq '');
+    ${$options{perf}}->{unit} = '%';
+
+    if (!defined(${$options{perf}}->{max}) || ${$options{perf}}->{max} eq '') {
+        ${$options{perf}}->{value} = 0;
+        return;
+    }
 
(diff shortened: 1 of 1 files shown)


garnier-quentin:
Sorry but i don't want to apply (change the value) if there are no max (for invert).

UrBnW:
Here we're in `percent` transform, not in `revert` ?

garnier-quentin:
I know. I have done a mistake :)

garnier-quentin:
You want to set 0% if you can percent because of max missing ?

UrBnW:
Yes, I compute % value on some devices, where max value comes from the device.
Unfortunately, sometimes, this max value has not properly been set by the device administrators, thus returning 0.
I then would like the % value to be 0, instead of the original value (which is then not a %).

UrBnW:
I think it's more correct / relevant...

garnier-quentin:
I prefer to add an attribute to the function.

UrBnW:
For completeness, here's my production example, without this PR :
```
--plugin=network::checkpoint::snmp::plugin --mode=connections --units=% --extend-perfdata='(.*),$1_prct,percent()'
OK: Connections: 9 | 'connections'=9con;;;0; 'connections_prct'=9con;;;0;
```

With this PR :
```
--plugin=network::checkpoint::snmp::plugin --mode=connections --units=% --extend-perfdata='(.*),$1_prct,percent()'
OK: Connections: 9 | 'connections'=9con;;;0; 'connections_prct'=0%;;;0;
```

UrBnW:
> I prefer to add an attribute to the function.

Dont bother / waste time with this then, I'll make a fix in the plugin itself.

PR #2002, opened 2020-05-14
fix(checkpoint) unknown if can't compute thresholds

UrBnW:
Hi,

This PR fixes the `connections` mode of `checkpoint` plugin.
If returns `UNKNOWN` in case it is called with `%` thresholds, but can't compute these thresholds, because of a missing max value.
User then knows he must properly configure max value at device level, before playing with `%` thresholds.

Thx 👍


UrBnW:
Thank you Quentin 👍

PR #2007, opened 2020-05-19
enh(ssh) Use BatchMode

UrBnW:
Hi,

Here's a PR to use SSH BatchMode by default.

Thank you 👍


garnier-quentin:
i need to use the new ssh backend system (which get that option). Tell me the plugin you use.

UrBnW:
`os::aix::local::plugin`, `errpt` mode.

Issue #1989, opened 2020-05-08
Add --use-new-perfdata option (and new labels) to various plugins


UrBnW:
Hi,

Could you add the `--use-new-perfdata` option to get the new labels (with their `instance#` prefix if relevant), to the following plugins please ?

- `network::checkpoint::snmp` : see PR #2008
- `network::citrix::netscaler::snmp` : see PR #2034 / #2071
- `network::fortinet::fortigate::snmp` : see PR #2035
- `network::watchguard::snmp` : see PR #2037
- `cloud::microsoft::office365::*` : see PR #2042
- `apps::citrix::local` : see PR #2077

Many thanks :+1:

UrBnW:
Closing this as every requested item now has its PR. Thx 👍

PR #2012, opened 2020-05-25
enh(checkpoint) add new labels

UrBnW:
Hi,

Let's add some new labels for Checkpoint devices.
This helps #1989.
And also relies on #2008 for the CPU new labels.

Thx 👍


UrBnW:
Of course feel free if the name of the new labels are not correct 👍

garnier-quentin:
You can test it.

PR #2035, opened 2020-06-04
enh(fortigate) Add newlabels

UrBnW:
Hi,

Let's add some new labels for Fortigate devices.
This helps #1989.
Of course feel free if the name of the new labels are not correct !

Thx 👍


UrBnW:
There's an issue with the `vpn` mode, where perfdata unit is `b/s`, while OIDs return `Bytes`.
I made no correction, as I'm not sure what's the unit you prefer (seems to be `b/s`, but a few modes still use `B/s`). Thx !

UrBnW:
While here, correct a unit issue in the `vpn` mode, where traffic unit is set to bits (`unit => 'b/s'`), while OIDs return bytes.
Let's then properly move to bits per second, as in (almost ?) any other traffic mode.

PR #2037, opened 2020-06-04
enh(watchguard) Add newlabels

UrBnW:
Hi,

Let's add some new labels for Watchguard devices.
This helps #1989.
Of course feel free if the name of the new labels are not correct !

Thx 👍

### DIFF
--- a/network/watchguard/snmp/mode/cpu.pm
+++ b/network/watchguard/snmp/mode/cpu.pm
@@ -33,7 +33,8 @@ sub set_counters {
     ];
     
     $self->{maps_counters}->{global} = [
-        { label => '1min', set => {
+        # I did not put '1m' in the label below, to mimic snmp_standard::mode::cpu
+        { label => '1min', nlabel => 'cpu.utilization.percentage', set => {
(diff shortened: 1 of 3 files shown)


UrBnW:
There's an issue with the `policyusage` and `system` modes, where perfdata unit is `b/s`, while OIDs return `Bytes`.
I made no correction, as I'm not sure what's the unit you prefer (seems to be `b/s`, but a few modes still use `B/s`). Thx !

UrBnW:
While here, correct a unit issue in the `policyusage` and `system` modes, where traffic unit is set to bits (`unit => 'b/s'`), while OIDs return bytes.

UrBnW:
And while here, remove the `diff => 1` option from the `total-connections` counter, `policy-usage` mode.
Seems like a typo, but I may be wrong...

UrBnW:
Just reverted no more needed CPU new-labels, because CPU mode gets fully updated in #2069.

garnier-quentin:
You can test it

UrBnW:
Works perfectly, thx 👍

PR #2038, opened 2020-06-04
fix(http::expectedcontent) support PATCH & PUT

### LINKED ISSUE #2036: apps::protocols::http::plugin - method PATCH / PUT with DATA
Hi,

I'm trying to monitor a web service with apps::protocols::http::plugin using --method=PATCH and a DATA content (--data=xxxxx).

Reading my Nginx logs, I see that the request is sent with the POST method whenever I set the --data parameter, without the --data, method sent to the web server is the good one PATCH.

I have the same issue using PUT method.

Same request using curl gave me the expected result.

Kind regards,

Bertrand.

UrBnW:
Hi,

Should close #2036.

Thx 👍

### DIFF
--- a/apps/protocols/http/mode/expectedcontent.pm
+++ b/apps/protocols/http/mode/expectedcontent.pm
@@ -147,7 +147,8 @@ sub load_request {
     my ($self, %options) = @_;
 
     $self->{options_request} = {};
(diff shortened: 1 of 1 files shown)


BertrandJO:
Thx everybody

PR #2042, opened 2020-06-05
enh(o365) Many enhancements

UrBnW:
Hi,

Here's a rework of Office365 plugins.

- I add some new labels (this helps #1989).
- I rename some counters to better correspond to what the API calls return, and to avoid having identical counters over different modes.
- I add a few new counters.
- I update some instance labels to make things proper / easier.
- I rename `onedrive` `usage` mode to `site-usage` (as it is with `sharepoint`).
- I add the `onedrive` `users-activity` mode (as we have with `sharepoint`).
- I now use daily data to compute the active counters, to have more relevant numbers than the 7-days aggregated ones.
- I add a `perfdate` perfdata counter set to the _Report Refresh Date_, so that it can be used by Stream Connector Scripts to properly set the data timestamp (see https://github.com/centreon/centreon-stream-connector-scripts/pull/23).

Thank you 👍

### DIFF
--- a/cloud/microsoft/office365/exchange/mode/emailactivity.pm
+++ b/cloud/microsoft/office365/exchange/mode/emailactivity.pm
@@ -189,25 +189,28 @@ sub manage_selection {
     $self->{global} = { send_count => 0, receive_count => 0 , read_count => 0 };
     $self->{users} = {};
 
-    my $results = $options{custom}->office_get_exchange_activity();
+    my $results = $options{custom}->office_get_exchange_activity(param => "period='D7'");
+    my $results_daily = [];
+    if (scalar(@{$results})) {
+       $self->{active}->{report_date} = @{$results}[0]->{'Report Refresh Date'};
+       $results_daily = $options{custom}->office_get_exchange_activity(param => "date=" . $self->{active}->{report_date});
+    }
 
-    foreach my $user (@{$results}) {
+    foreach my $user (@{$results}, @{$results_daily}) {
         # Let's lc the instance label to make metrics "clean"...
         $user->{'User Principal Name'} = lc($user->{'User Principal Name'});
-
-        $self->{active}->{report_date} = $user->{'Report Refresh Date'} if ($self->{active}->{report_date} eq '');
 
         if (defined($self->{option_results}->{filter_user}) && $self->{option_results}->{filter_user} ne '' &&
             $user->{'User Principal Name'} !~ /$self->{option_results}->{filter_user}/) {
             $self->{output}->output_add(long_msg => "skipping  '" . $user->{'User Principal Name'} . "': no matching filter name.", debug => 1);
             next;
         }
-    
-        $self->{active}->{total}++;
 
-        if (!defined($user->{'Last Activity Date'}) || $user->{'Last Activity Date'} eq '' ||
-            ($user->{'Last Activity Date'} ne $user->{'Report Refresh Date'})) {
-            $self->{output}->output_add(long_msg => "skipping '" . $user->{'User Principal Name'} . "': no activity.", debug => 1);
+        if ($user->{'Report Period'} != 1) {
+            if (!defined($user->{'Last Activity Date'}) || ($user->{'Last Activity Date'} ne $self->{active}->{report_date})) {
+                $self->{output}->output_add(long_msg => "skipping '" . $user->{'User Principal Name'} . "': no activity.", debug => 1);
+            }
+            $self->{active}->{total}++;
             next;
         }
 
@@ -230,7 +233,7 @@ __END__
 
 =head1 MODE
 
-Check email activity (reporting period over the last 7 days).
+Check email activity (reporting period over the last refreshed day).
 
 (See link for details about metrics :
 https://docs.microsoft.com/en-us/office365/admin/activity-reports/email-activity?view=o365-worldwide)
--- a/cloud/microsoft/office365/exchange/mode/emailactivity.pm
+++ b/cloud/microsoft/office365/exchange/mode/emailactivity.pm
@@ -36,7 +36,7 @@ sub custom_active_perfdata {
     }
 
     $self->{result_values}->{report_date} =~ /^([0-9]{4})-([0-9]{2})-([0-9]{2})$/;
-    $self->{output}->perfdata_add(label => 'perfdate', value => timegm(0,0,0,$3,$2-1,$1-1900));
+    $self->{output}->perfdata_add(label => 'perfdate', value => timelocal(0,0,0,$3,$2-1,$1-1900));
 
     $self->{output}->perfdata_add(label => 'active_users', nlabel => 'exchange.users.active.count',
                                   value => $self->{result_values}->{active},
--- a/cloud/microsoft/office365/exchange/mode/emailactivity.pm
+++ b/cloud/microsoft/office365/exchange/mode/emailactivity.pm
@@ -34,11 +34,11 @@ sub custom_active_perfdata {
         $total_options{cast_int} = 1;
     }
 
-    $self->{output}->perfdata_add(label => 'active_mailboxes',
+    $self->{output}->perfdata_add(label => 'active_users', nlabel => 'exchange.users.active.count',
                                   value => $self->{result_values}->{active},
                                   warning => $self->{perfdata}->get_perfdata_for_output(label => 'warning-' . $self->{label}, %total_options),
                                   critical => $self->{perfdata}->get_perfdata_for_output(label => 'critical-' . $self->{label}, %total_options),
-                                  unit => 'mailboxes', min => 0, max => $self->{result_values}->{total});
+                                  unit => 'users', min => 0, max => $self->{result_values}->{total});
 }
 
 sub custom_active_threshold {
@@ -58,7 +58,7 @@ sub custom_active_threshold {
 sub custom_active_output {
     my ($self, %options) = @_;
 
-    my $msg = sprintf("Active mailboxes on %s : %d/%d (%.2f%%)",
+    my $msg = sprintf("Active users on %s : %d/%d (%.2f%%)",
                         $self->{result_values}->{report_date},
                         $self->{result_values}->{active},
                         $self->{result_values}->{total},
@@ -80,13 +80,13 @@ sub custom_active_calc {
 sub prefix_global_output {
     my ($self, %options) = @_;
     
-    return "Total (active mailboxes) ";
+    return "Total (active users) ";
 }
 
 sub prefix_mailbox_output {
     my ($self, %options) = @_;
     
-    return "Mailbox '" . $options{instance_value}->{name} . "' ";
+    return "User '" . $options{instance_value}->{name} . "' ";
 }
 
 sub set_counters {
@@ -95,11 +95,11 @@ sub set_counters {
     $self->{maps_counters_type} = [
         { name => 'active', type => 0 },
         { name => 'global', type => 0, cb_prefix_output => 'prefix_global_output' },
-        { name => 'mailboxes', type => 1, cb_prefix_output => 'prefix_mailbox_output', message_multiple => 'All email activity are ok' },
+        { name => 'users', type => 1, cb_prefix_output => 'prefix_mailbox_output', message_multiple => 'All email activity are ok' },
     ];
     
     $self->{maps_counters}->{active} = [
-        { label => 'active-mailboxes', set => {
+        { label => 'active-users', set => {
                 key_values => [ { name => 'active' }, { name => 'total' }, { name => 'report_date' } ],
                 closure_custom_calc => $self->can('custom_active_calc'),
                 closure_custom_output => $self->can('custom_active_output'),
@@ -109,7 +109,7 @@ sub set_counters {
         },
     ];
     $self->{maps_counters}->{global} = [
-        { label => 'total-send-count', set => {
+        { label => 'total-send-count', nlabel => 'exchange.users.emails.sent.total.count', set => {
                 key_values => [ { name => 'send_count' } ],
                 output_template => 'Send Count: %d',
                 perfdatas => [
@@ -118,7 +118,7 @@ sub set_counters {
                 ],
             }
         },
-        { label => 'total-receive-count', set => {
+        { label => 'total-receive-count', nlabel => 'exchange.users.emails.received.total.count', set => {
                 key_values => [ { name => 'receive_count' } ],
                 output_template => 'Receive Count: %d',
                 perfdatas => [
@@ -127,7 +127,7 @@ sub set_counters {
                 ],
             }
         },
-        { label => 'total-read-count', set => {
+        { label => 'total-read-count', nlabel => 'exchange.users.emails.read.total.count', set => {
                 key_values => [ { name => 'read_count' } ],
                 output_template => 'Read Count: %d',
                 perfdatas => [
@@ -137,8 +137,8 @@ sub set_counters {
             }
         },
     ];
-    $self->{maps_counters}->{mailboxes} = [
-        { label => 'send-count', set => {
+    $self->{maps_counters}->{users} = [
+        { label => 'send-count', nlabel => 'exchange.users.emails.sent.count', set => {
                 key_values => [ { name => 'send_count' }, { name => 'name' } ],
                 output_template => 'Send Count: %d',
                 perfdatas => [
@@ -147,7 +147,7 @@ sub set_counters {
                 ],
             }
         },
-        { label => 'receive-count', set => {
+        { label => 'receive-count', nlabel => 'exchange.users.emails.received.count', set => {
                 key_values => [ { name => 'receive_count' }, { name => 'name' } ],
                 output_template => 'Receive Count: %d',
                 perfdatas => [
@@ -156,7 +156,7 @@ sub set_counters {
                 ],
             }
         },
-        { label => 'read-count', set => {
+        { label => 'read-count', nlabel => 'exchange.users.emails.read.count', set => {
                 key_values => [ { name => 'read_count' }, { name => 'name' } ],
                 output_template => 'Read Count: %d',
                 perfdatas => [
@@ -174,7 +174,7 @@ sub new {
     bless $self, $class;
     
     $options{options}->add_options(arguments => {
-        "filter-mailbox:s"      => { name => 'filter_mailbox' },
+        "filter-user:s"         => { name => 'filter_user' },
         "units:s"               => { name => 'units', default => '%' },
         "filter-counters:s"     => { name => 'filter_counters', default => 'active|total' }, 
     });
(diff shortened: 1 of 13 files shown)


UrBnW:
I just finished to work on this, I added various enhancements (see first post).
Properly tested, it's ready to land.
Thank you 👍

Sims24:
Hi @UrBnW thanks for this wonderfull work. I will take a deeper look asap.

UrBnW:
Thank you very much @Sims24 for taking the time to review it 👍
Of course, feel free if needed, if some of the changes are not understandable / clear enough, or whatever...

cloud/microsoft/office365/exchange/mode/emailactivity.pm
```diff
@@ -187,36 +191,43 @@ sub manage_selection {
     
     $self->{active} = { active => 0, total => 0, report_date => '' };
     $self->{global} = { send_count => 0, receive_count => 0 , read_count => 0 };
-    $self->{mailboxes} = {};
+    $self->{users} = {};
+
+    my $results = $options{custom}->office_get_exchange_activity(param => "period='D7'");
```
Sims24, line 196: @UrBnW why do you keep an hardcoded period here?

cloud/microsoft/office365/exchange/mode/emailactivity.pm
```diff
@@ -24,6 +24,7 @@ use base qw(centreon::plugins::templates::counter);
 
 use strict;
 use warnings;
+use Time::Local;
```
Sims24, line 27: I'm not sure about this. 

I'm not very familiar with this lib and the timegm() function ... My understanding is that we will always get UTC instead of local TZ? Can't we imagine give the ability for the user to choose the TZ?

cloud/microsoft/office365/exchange/mode/emailactivity.pm
UrBnW: The various API calls require to set either [period or date](https://docs.microsoft.com/en-us/graph/api/reportroot-getemailactivityuserdetail?view=graph-rest-1.0).
Period is D7, D30, D90, or D180.
Date is YYYY-MM-DD, for which you would like to view the users who performed any activity.

So, using period `D7` in the first API call mainly allows us to get 2 things :
- the number of inactive objects ;
- the date the report has been generated, which will be used as the date parameter to the second API call.

We can't directly do the second call, as we don't know the last refresh date (which varies and is sometimes a few days ago).
And this second call does not return inactive objects information.

cloud/microsoft/office365/exchange/mode/emailactivity.pm
UrBnW: Sorry for my late answer on this @Sims24.

You're right, `timegm` will generate a UTC timestamp.
Broker itself uses UTC timestamps in the events.
We then must use UTC in the `perfdate` timestamp we generate (the _Report Refresh Date_) to replace the default one (the "now" timestamp).

UrBnW: Sorry for my late answer on this @Sims24.

You're right, `timegm` will generate a UTC timestamp.
And after digging into this again, timestamps used in the events (by the broker etc...) are local timestamps.
We then must use local time in the `perfdate` timestamp we generate (the _Report Refresh Date_) to replace the default one (the "now" timestamp).
Let me push a fix accordingly.

UrBnW: Fix pushed, thank you for this 👍

UrBnW:
Some news regarding this review perhaps @Sims24 ?
Many thanks !

cloud/microsoft/office365/exchange/mode/emailactivity.pm
Sims24: This is under investigation on our side. ANy feedback about double-call overhead? Is this fast enough to satisfy everybody? Do you have a comparison of what it implies in terms of exec time.

Sims24:
@UrBnW I add Colin as he is the one who wrote it initially. We are discussing with @cgagnaire and @garnier-quentin and it may take some time for us to agree on all the point 😹

cloud/microsoft/office365/exchange/mode/emailactivity.pm
UrBnW: Yes, I run all these modified plugins since weeks now.
Their execution time varies between one and two seconds.
Seems largely acceptable, keeping in mind that it's for about 10 services overall, and that the check frequency can be hour-based, instead of (generally) minute-based, as the reported data is daily-based, not continuously refreshed.
I run them hourly, everything's fine 👍

UrBnW:
Hi @Sims24, many thanks for your feedback 👍
Of course feel free if needed.
I would really be glad if behaviour introduced here could be kept, as it allows to grab the most accurate data possible, both global and per-user.

Leading to nice graphs / metrics 😉

Screenshot 2020-07-29 at 10 37 24

Thank you again !

Sims24:
@UrBnW thanks for sharing more informations about this. 

We wonder also why would user be more relevant that mailbox in the perfdata label? It's breaking things and we do feel comfortable that a mailbox is always linked to a user. Wdyt ?

UrBnW:
`emailactivity` uses `users` because it better suits what API itself returns, "users" :
https://docs.microsoft.com/en-us/microsoft-365/admin/activity-reports/email-activity?view=o365-worldwide
_The table shows you a breakdown of the email activities at the per-user level. This shows all users that have an Exchange product assigned to them and their email activities._

And `mailboxusage` uses `mailboxes`, as what returns API :
https://docs.microsoft.com/en-us/office365/admin/activity-reports/mailbox-usage?view=o365-worldwide
_The table shows you a breakdown of mailbox usage at the per-user level._

It also better conforms with other modes, where `*activity` are all user-related, `*usage` are "things"-related (sites, mailboxes, devices...).

This makes all the o365 plugins suite more consistent :)

cloud/microsoft/office365/exchange/mode/emailactivity.pm
```diff
@@ -34,11 +35,14 @@ sub custom_active_perfdata {
         $total_options{cast_int} = 1;
     }
 
-    $self->{output}->perfdata_add(label => 'active_mailboxes',
+    $self->{result_values}->{report_date} =~ /^([0-9]{4})-([0-9]{2})-([0-9]{2})$/;
+    $self->{output}->perfdata_add(label => 'perfdate', value => timelocal(0,0,12,$3,$2-1,$1-1900));
```
cgagnaire, line 39: Can you use a more clear metric name like 'report_date', and maybe add a new label also.

cloud/microsoft/office365/exchange/mode/emailactivity.pm
```diff
@@ -187,36 +191,43 @@ sub manage_selection {
...
+    my $results_daily = [];
+    if (scalar(@{$results})) {
+       $self->{active}->{report_date} = @{$results}[0]->{'Report Refresh Date'};
+       $results_daily = $options{custom}->office_get_exchange_activity(param => "date=" . $self->{active}->{report_date});
+    }
 
-    my $results = $options{custom}->office_get_exchange_activity();
+    foreach my $user (@{$results}, @{$results_daily}) {
+        # Let's lc the instance label to make metrics "clean"...
+        $user->{'User Principal Name'} = lc($user->{'User Principal Name'});
```
cgagnaire, line 205: Let's not do that. Eveything has a purpose in life, and you can use change-perfdata mecanism to meet your needs.

cloud/microsoft/office365/exchange/mode/emailactivity.pm
```diff
@@ -187,36 +191,43 @@ sub manage_selection {
...
-        $self->{active}->{total}++;
 
-        if (!defined($mailbox->{'Last Activity Date'}) || $mailbox->{'Last Activity Date'} eq '' ||
-            ($mailbox->{'Last Activity Date'} ne $mailbox->{'Report Refresh Date'})) {
-            $self->{output}->output_add(long_msg => "skipping '" . $mailbox->{'User Principal Name'} . "': no activity.", debug => 1);
+        if ($user->{'Report Period'} != 1) {
+            if (!defined($user->{'Last Activity Date'}) || ($user->{'Last Activity Date'} ne $self->{active}->{report_date})) {
+                $self->{output}->output_add(long_msg => "skipping '" . $user->{'User Principal Name'} . "': no activity.", debug => 1);
+            }
+            $self->{active}->{total}++;
```
cgagnaire, line 217: This mean it excludes every accounts that never had an activity?
That doesn't suite me. I prefer adding another metric like active/inactive/total.

cloud/microsoft/office365/exchange/mode/mailboxusage.pm
```diff
@@ -245,25 +249,31 @@ sub manage_selection {
     $self->{global} = { storage_used_active => 0, storage_used_inactive => 0 };
     $self->{mailboxes} = {};
 
-    my $results = $options{custom}->office_get_exchange_mailbox_usage();
+    my $results = $options{custom}->office_get_exchange_mailbox_usage(param => "period='D7'");
+    my $results_daily = [];
+    if (scalar(@{$results})) {
+       $self->{active}->{report_date} = @{$results}[0]->{'Report Refresh Date'};
+       #$results_daily = $options{custom}->office_get_exchange_mailbox_usage(param => "date=" . $self->{active}->{report_date});
```
cgagnaire, line 256: Why not here?

cloud/microsoft/office365/onedrive/mode/siteusage.pm
```diff
@@ -262,12 +275,21 @@ sub manage_selection {
...
-    my $results = $options{custom}->office_get_onedrive_usage();
+    my $results = $options{custom}->office_get_onedrive_usage(param => "period='D7'");
+    my $results_daily = [];
+    if (scalar(@{$results})) {
+       $self->{active}->{report_date} = @{$results}[0]->{'Report Refresh Date'};
+       $results_daily = $options{custom}->office_get_onedrive_usage(param => "date=" . $self->{active}->{report_date});
+    }
+
+    foreach my $site (@{$results}, @{$results_daily}) {
+        # As it's used as the instance label, let's keep the URL as short as possible, removing its domain name...
```
cgagnaire, line 290: Same same

cloud/microsoft/office365/onedrive/mode/siteusage.pm
```diff
@@ -173,36 +177,45 @@ sub set_counters {
...
-                    { label => 'storage_used_inactive', value => 'storage_used_inactive', template => '%d',
+                    { label => 'total_usage_inactive', value => 'storage_used_inactive', template => '%d',
                       min => 0, unit => 'B' },
                 ],
             }
         },
-        { label => 'total-file-count', set => {
-                key_values => [ { name => 'file_count' } ],
+        { label => 'total-file-count-active', nlabel => 'onedrive.sites.active.files.total.count', set => {
+                key_values => [ { name => 'file_count_active' } ],
```
cgagnaire, line 201: Something's wrong between `file_count_active` and `active_file_count`.
They may be a "active" (for sites) missing somewhere.

cloud/microsoft/office365/skype/mode/devicesusage.pm
```diff
@@ -58,7 +62,7 @@ sub custom_active_threshold {
 sub custom_active_output {
     my ($self, %options) = @_;
 
-    my $msg = sprintf("Active users on %s : %d/%d (%.2f%%)",
+    my $msg = sprintf("Active devices on %s : %d/%d (%.2f%%)",
```
cgagnaire, line 65: Can we keep the active users? And add an "active devices" instead.

cloud/microsoft/office365/teams/mode/devicesusage.pm
```diff
@@ -58,7 +62,7 @@ sub custom_active_threshold {
 sub custom_active_output {
     my ($self, %options) = @_;
 
-    my $msg = sprintf("Active users on %s : %d/%d (%.2f%%)",
+    my $msg = sprintf("Active devices on %s : %d/%d (%.2f%%)",
```
cgagnaire, line 65: Same as for Skype.

UrBnW:
Hi @cgagnaire, thank you for your review, let me answer it :)

cloud/microsoft/office365/exchange/mode/emailactivity.pm
```diff
@@ -58,7 +62,7 @@ sub custom_active_threshold {
 sub custom_active_output {
     my ($self, %options) = @_;
 
-    my $msg = sprintf("Active mailboxes on %s : %d/%d (%.2f%%)",
+    my $msg = sprintf("Active users on %s : %d/%d (%.2f%%)",
```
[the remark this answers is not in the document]
UrBnW: Look at the API calls used :

`emailactivity` mode : `getEmailActivityUserDetail` [doc](https://docs.microsoft.com/en-us/graph/api/reportroot-getemailactivityuserdetail?view=graph-rest-1.0) states : _Get details about email activity **users** have performed. Specifies the date for which you would like to view the **users who performed** any activity._
`mailboxusage` mode : `getMailboxUsageDetail` [doc](https://docs.microsoft.com/en-us/graph/api/reportroot-getmailboxusagedetail?view=graph-rest-1.0) states : _Get details about **mailbox** usage._

So the `emailactivity` is about users, the `mailboxusage` is about mailboxes.
I wonder whether or not users could send mails from more than one mailbox. Think for example about shared mailboxes, for example admin@mycorp.xyz, where every admin answers to received mails with his own name (we then see "_sent by john@mycorp.xyz_" in the received mail).
I then assume this can be the case. This may be why `getEmailActivityUserDetail` API call is about users, not about mailboxes.

In addition to this, I think it's more proper / relevant to keep the unit as stated in the API documentation. Instead of making asumptions changing it :)

It's also more consistent with other modes : all other `*activity` modes are about users (`onedrive`, `sharepoint`, `skype`, `teams`). And the `*usage` modes are respectively about sites (`onedrive`, `sharepoint`) and devices (`skype`, `teams`).
This makes the whole thing properly / easily readable.

I think it's not a big problem to change one unit here (even more into the documented one), as part of a big rework of the plugin. Some Centreon Plugins' changes are sometimes breaking changes, even sometimes unit changes...
This answers your second point [here](https://github.com/centreon/centreon-plugins/pull/2042#pullrequestreview-457548963).

cloud/microsoft/office365/exchange/mode/emailactivity.pm
UrBnW: Office365 API reports daily data, so at least from the day before, sometimes from several days ago, depending on how fast Microsoft generates reports at this moment...
This `perfdate` metric is then set to the _Report Refresh Date_ from the API result.
The only purpose of this metric is to be handled by Stream Connector Scripts to properly set the data timestamp (see https://github.com/centreon/centreon-stream-connector-scripts/pull/23). So that data is sent to InfluxDB (for instance) with the proper report date, not the request date... Leading to accurate graphs as shown [above](https://github.com/centreon/centreon-plugins/pull/2042#issuecomment-665525294).
`perfdate` then sounds ideal for this purpose, as the _perfdata_ date.
As a "meta" metric, it does not need its own graph as any other proper metric does, thus does not need a new label.
Btw, note that InfluxDB Stream Connector Script rejects metrics which are not new-label named, which is nice, `perfdate` will not unnecessarily reach InfluxDB.
Finally, this `perfdate` metric could be used in any other mode facing the same non-realtime data behaviour.

cloud/microsoft/office365/exchange/mode/emailactivity.pm
UrBnW: No, I think you did not catch the algorithm principle. Let me explain.

For every mode, we do 2 requests :
- a `period` request (`D7`, so 7 days), which returns the number of active and inactive users (or sites etc...), aggregated metrics over the period, and the _Report Refresh Date_ ;
- a `date` request, at the date of _Report Refresh Date_ received from the previous request, which returns metrics for the active users (or sites...) at this specific date.

Thanks to the first call, we then can still compute the number of active and inactive users, as before.
And thanks to the second call, we compute much more precise per-user (or per-site...) metrics, because not aggregated over 7 days, but over 1 precise day only. Which is much more relevant, especially on graphs...

So in this piece of code (longer quote) :
https://github.com/centreon/centreon-plugins/blob/37172649f9380818b0f711516ad35be94319528b/cloud/microsoft/office365/exchange/mode/emailactivity.pm#L213-L219
We compute the total number of accounts, thanks to the `period` results (`Report Period == 7`).
And, thanks to the `next` instruction, we skip the code below, which concerns active users (or sites...) only, having their accurate data in `date` results (`Report Period == 1`).

So nothing wrong / missing here :)

About the 2 calls scheme. We can't directly do the second (`date`) call, as we don't know the last refresh date (which varies and is sometimes a few days ago). And in addition, this second call does not return inactive objects information. We then need both of them. But does not impact performances as explained [above](https://github.com/centreon/centreon-plugins/pull/2042#discussion_r462133126).
This answers your first point [here](https://github.com/centreon/centreon-plugins/pull/2042#pullrequestreview-457548963).

cloud/microsoft/office365/exchange/mode/mailboxusage.pm
UrBnW: Because the [`getMailboxUsageDetail` API call](https://docs.microsoft.com/en-us/graph/api/reportroot-getmailboxusagedetail?view=graph-rest-1.0) only support _period_ request, not _date_ request. It's the only one like this.

cloud/microsoft/office365/onedrive/mode/siteusage.pm
UrBnW: `onedrive.sites.active.files.total.count` is the number of files in active sites.
`onedrive.sites.files.active.total.count` is the number of active files (so in active sites).
We could rename the later into `onedrive.sites.active.files.active.total.count`, not sure though the additional `active` is needed here...
Up to you.
Same thing then for the last 3 global sharepoint metrics.

cloud/microsoft/office365/skype/mode/devicesusage.pm
UrBnW: We already have the active / inactive users metrics in `useractivity` mode.
I then thought useless to duplicate the metric, which would then be visible 2 times at different places in Centreon, Stream Connector Scripts' targets such as InfluxDB etc... There's no need to have 2 different metrics pointing to the same thing...
I also found more interesting to report devices, in a `deviceusage` mode, properly separating things, making again the whole thing more readable.

cloud/microsoft/office365/teams/mode/devicesusage.pm
UrBnW: See [above](https://github.com/centreon/centreon-plugins/pull/2042#discussion_r464733505).

cloud/microsoft/office365/exchange/mode/emailactivity.pm
UrBnW: Cleaning removed.

cloud/microsoft/office365/onedrive/mode/siteusage.pm
UrBnW: As [above](https://github.com/centreon/centreon-plugins/pull/2042#discussion_r464734744), instances cleaning remove for all other modes in https://github.com/centreon/centreon-plugins/pull/2042/commits/181d60f5162147fc4b7a466057fdddcc20476906.

UrBnW:
Hi @cgagnaire, thank you very much for your review !
I tried to answer every point precisely.
Feel free if you have any other question 👍

cloud/microsoft/office365/exchange/mode/emailactivity.pm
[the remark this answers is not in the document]
UrBnW: And a real-life example :
I have an environment where `emailactivity` reports 380 **users**, and `mailboxusage` reports 90 **mailboxes**.
So total different numbers.
And, depending on the days, number of active **users** different than number of active **mailboxes**.
These are then 2 real different things.

cgagnaire (approved):

UrBnW:
Many thanks Colin, really glad, good news 👍

UrBnW:
Fantastic :)
Thank you Simon 👍

PR #2061, opened 2020-06-16
fix(idrac) "other" status is WARNING

UrBnW:
Hi,

As it is already in `globalstatus` mode, let's make the `other` status return `WARNING` (instead of `UNKNOWN` currently).
It's then more consistent.

Same thing with `nonRecoverable` status, let's make it consistent over the different modes...

Some thoughts regarding `unknown` status set to an `UNKNOWN` alert level.
It's true that `unknown` = `UNKNOWN` :)
But here, the device properly responded, with an `unknown` status, shouldn't we consider this as a `WARNING` level ? Could be a proper issue.
I thought that `UNKNOWN` was rather meant for communication errors or inconsistent results / things.

Feel free if I'm wrong, and thank you for your feedback 👍

### DIFF
--- a/hardware/server/dell/idrac/snmp/mode/globalstatus.pm
+++ b/hardware/server/dell/idrac/snmp/mode/globalstatus.pm
@@ -173,12 +173,12 @@ Can used special variables like: %{status}
 
 =item B<--warning-status>
 
-Set warning threshold for status (Default: '%{status} =~ /nonRecoverable|non critical|other/i').
+Set warning threshold for status (Default: '%{status} =~ /non critical|other/i').
 Can used special variables like: %{status}
 
 =item B<--critical-status>
 
-Set critical threshold for status (Default: '%{status} =~ /^critical/i').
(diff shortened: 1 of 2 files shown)


UrBnW:
Same thing with `nonRecoverable` status, let's make it consistent over the different modes...

garnier-quentin:
You change nonRecoverable also...

garnier-quentin:
I wont change it i think. There are options to overload it.

UrBnW:
> You change nonRecoverable also...

Yes to make it consistent, I added a note to my first post.
You want `other` only ?

garnier-quentin:
I wont change it i think. UNKNOWN for 'other' seems ok for me. But you can overload it to WARNING with options.

UrBnW:
OK, that was to make things consistent over the whole `idrac` plugin, but of course yes they can be overloaded.
Up to you then :)

garnier-quentin:
I will quote De Gaulle: 
```
Comment voulez-vous gouverner un pays où il existe 258 variétés de fromage ?
```

;)

PR #2090, opened 2020-07-05
enh(whatchguard) Properly report 0 tunnel

UrBnW:
Hi,

It's a normal situation, from time to time, to have 0 VPN / IPsec tunnel, depending whether or not there are connected users.
In this case, let's then properly report `OK: Total Tunnels: 0`, instead of returning an `UNKNOWN` status, which is not really expected here.

Thx 👍


UrBnW:
Thank you 👍

PR #2102, opened 2020-07-11
enh(hardware) Allow to use regex threshold-overload

UrBnW:
Hi,

This PR enhances hardware template, allowing to use regex in threshold-overload section.
Useful when one wants to set custom thresholds on a lot of different devices / components at the same time.
Order of multiple options is kept, so that we can use both plain and regex thresholds at the same time, in a consistent manner.

Thank you 👍

### DIFF
--- a/centreon/plugins/templates/hardware.pm
+++ b/centreon/plugins/templates/hardware.pm
@@ -166,7 +166,7 @@ sub check_options {
         }
     }
 
-    $self->{overload_th} = {};
+    $self->{overload_th} = ();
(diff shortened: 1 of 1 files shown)


garnier-quentin:
Need to think about it because it could change the old configuration for users.

UrBnW:
Thank you for your feedback.
The verification I did is searching for all `regexp_threshold_overload_check_section_option` values.
They all satisfy the `^[a-z.]+$ ` regex.
It should then not break current custom thresholds, but perhaps I missed something :)

PR #2113, opened 2020-07-21
x509/validity: Workaround ignored timeout

julthomas:
Altough the timeout option is defined in getopts, it was not used.

Also, when added to IO::Socket::SSL->new(), the Timeout option gets
ignored in out environment, so we workaround it with an ALRM
sighandler.

### DIFF
--- a/apps/protocols/x509/mode/validity.pm
+++ b/apps/protocols/x509/mode/validity.pm
@@ -86,6 +86,15 @@ sub check_options {
 sub run {
     my ($self, %options) = @_;
 
+    # Workaround for ignored timeout. When the Timeout option is added to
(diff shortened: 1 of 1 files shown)


UrBnW:
This is due to `IO-Socket-SSL` version, too old on CentOS 7.
[Update it](https://metacpan.org/pod/IO::Socket::SSL), and it will solve your issue (you would also face with any other SSL/HTTPS plugin and its `--timeout` option).
I have `2.066` installed on my pollers, but last `2.068` should also be OK, no reason.

There's also a `--global-timeout` option which should help here.

garnier-quentin:
You should use the global timeout option.

julthomas:
OK thank you, looks like that option `--global-timeout` wasn't in the `--help`

PR #2114, opened 2020-07-21
x509/validity: Fix wrong exit status, some CRITICAL should be UNKNOWN

julthomas:
The real purpose of this check is testing the certificate,
so the following errors should return an UNKNOWN status error:
- connection failure (eg: timeout)
- failure to retrieve the server certificate


garnier-quentin:
The validity mode is deprecated. Please use the ```certificate``` mode. It would do the job

PR #2124, opened 2020-07-24
enh(watchguard) add some global counters

UrBnW:
Hi,

Let's add some Watchguard VPN global counters.

Thx 👍


garnier-quentin:
Could you test it ?

UrBnW:
It works, but I just faced a bug where an instance (tunnel) value is sometimes greater than the total...
I suspect the tunnels table not to have consistent IDs, which would then lead to this bug.
I'm troubleshooting to be sure.

garnier-quentin:
I see the issue. yes we cannot increment like that. i remove

UrBnW:
OK with last fixes, thx 👍

PR #2138, opened 2020-08-06
fix(windows) local rebootpending

UrBnW:
Hi,

Strangely enough I have Windows 2008 R2 hosts where OS build number fails to get retrieved.
Let's then simplify the method, we absolutely don't need this.

Thx 👍


UrBnW:
This properly fixes the bug on all servers where I encountered it 👍

PR #2159, opened 2020-08-18
Add eval on curl backend to accept any type of variable in curl-opt

alfredosilvestre-natixis:
Hello,

While testing Kerberos proxy authentication, I noticed that the "curl-opt" does not accept an array as an argument in the hash, only strings are translated since the argument is a string array.

This makes it that you can't use the [CURLOPT_RESOLVE](https://curl.haxx.se/libcurl/c/CURLOPT_RESOLVE.html) option to set an alternative name resolution.

I see that this option is used in line 297 to set the "http_peer_addr", however it does not work for the proxy name resolution, only for the host.

The need for this option is due to the Kerberos authentication process needing to have the SPN (proxyserver.mynetwork) in the --proxyurl option.

So if the proxy SPN is a VIP, we can't test the individual proxy servers with Kerberos by passing the server's IP, the SPN needs to be used and the resolve option as well so we can change the IP.

An example call in this case is the following:

`centreon_plugins.pl --plugin=apps::protocols::http::plugin --mode=response --hostname=www.google.com --port=80 --http-backend=curl --proxyurl=http://proxyserver.mynetwork:8080 --curl-opt="CURLOPT_PROXYAUTH => CURLAUTH_GSSNEGOTIATE" --curl-opt="CURLOPT_PROXYUSERPWD => ':'" --curl-opt="CURLOPT_USERAGENT => 'MyBrowser v0.1'" --curl-opt="CURLOPT_RESOLVE => ['proxyserver.mynetwork:8080:192.168.2.2']" --curl-opt="CURLOPT_SSL_VERIFYPEER => 0"`

Another option to solve this issue would be to add a new argument like the "http_peer_addr", for example "proxy_peer_addr" so we can set an alternative name resolution for the proxy.

### DIFF
--- a/centreon/plugins/backend/http/curl.pm
+++ b/centreon/plugins/backend/http/curl.pm
@@ -240,7 +240,7 @@ sub set_extra_curl_opt {
             }
         }
 
-        $self->curl_setopt(option => $fields->{key}, parameter => $fields->{value});
+        $self->curl_setopt(option => $fields->{key}, parameter => eval $fields->{value});
     }
 }
 


garnier-quentin:
I prefer to not use eval. Could you try like that ?
```
--curl-opt="CURLOPT_RESOLVE => [proxyserver.mynetwork:8080:192.168.2.2]" 
```

alfredosilvestre-natixis:
Doesn't work, the error is this:
UNKNOWN: curl setopt error: 'not an array

garnier-quentin:
Could you provide your command line ?

alfredosilvestre-natixis:
I did in the OP 😄 

The issue is that the "$fields->{value}" will always be a string, however, the "CURLOPT_RESOLVE" option expects an array.

/usr/lib/centreon/plugins/centreon_plugins.pl --plugin=apps::protocols::http::plugin --mode=response --hostname='www.google.com' --port='80' --proto='http' --http-backend=curl --proxyurl=http://proxyserver.mynetwork:8080 --curl-opt="CURLOPT_PROXYAUTH => CURLAUTH_GSSNEGOTIATE" --curl-opt="CURLOPT_PROXYUSERPWD => ':'" --curl-opt="CURLOPT_USERAGENT => 'MyBrowser v0.1'" --curl-opt="CURLOPT_RESOLVE => ['proxyserver.mynetwork:8080:192.168.2.2']" --curl-opt="CURLOPT_SSL_VERIFYPEER => 0"

garnier-quentin:
Could you try like that ?
```
--curl-opt="CURLOPT_PROXYAUTH => CURLAUTH_GSSNEGOTIATE" --curl-opt="CURLOPT_PROXYUSERPWD => :" --curl-opt="CURLOPT_USERAGENT => MyBrowser v0.1" --curl-opt="CURLOPT_RESOLVE => [proxyserver.mynetwork:8080:192.168.2.2]" --curl-opt="CURLOPT_SSL_VERIFYPEER => 0"
```

alfredosilvestre-natixis:
Same error:
UNKNOWN: curl setopt error: 'not an array at /usr/lib/centreon/plugins/centreon/plugins/backend/http/curl.pm line 149.

garnier-quentin:
It's working for me... With centos 7 and xubuntu 20.04. Do you use the last version ? What is your OS and perl-Net-Curl version ?

alfredosilvestre-natixis:
Ah, my bad, didn't realize you updated the plugin.

I confirm it is working with both:
--curl-opt="CURLOPT_RESOLVE => [proxyserver.mynetwork:8080:192.168.2.2]"
--curl-opt="CURLOPT_RESOLVE => ['proxyserver.mynetwork:8080:192.168.2.2']"

Red Hat Enterprise Linux Server release 7.8 (Maipo)
perl-Net-Curl-0.39-1
Centreon 19.04.18

Thanks for the update 😄

garnier-quentin:
Great! Thanks for the feedback

PR #2167, opened 2020-08-24
fix dell fastpath cpu 5sec thresholds

meni2029:
Small fix for warning and critical thresholds on Dell CPU 5sec. Thanks !


Sims24 (approved):
Thank you !

PR #2169, opened 2020-08-24
fix(nsclient) properly decode JSON

UrBnW:
Hi,

I faced some issue with `nsclient` plugin not properly decoding JSON containing special characters.
This PR solves this.

Thx 👍


UrBnW:
Thank you Quentin 👍

PR #2179, opened 2020-08-31
Verbose help & type insensitive filtering

Sims24:


garnier-quentin (approved):
It's ok. But there is a more flexible syntax for that: --filter-type='(?i)toto'

PR #2189, opened 2020-09-03
(Fix): missing () in the regex

pkriko:
'Cause the missing (), $1 was undef and the options doesn't work


UrBnW:
No, it is up to you to place the parenthesis wherever you want in your `--extracted-pattern` option.

For example : `--extracted-pattern=',Connections: ([0-9]+),' --critical-extracted=80`

garnier-quentin:
Agree with @UrBnW

PR #2202, opened 2020-09-08
fix(interfaces) Properly handle 0 values with --force-counters64

UrBnW:
Hi,

This PR fixes usage of `--force-counters64` option in `interfaces` mode when interfaces return `0` values.

Thx 👍


garnier-quentin:
Could you test it ?

UrBnW:
Works, but see #2207 for a last fix :)
Thx 👍

PR #2220, opened 2020-09-21
Fixed infinite loop on multiple page output

alfredosilvestre-natixis:
Hello,

On using this plugin when it returns multiple pages, it will stay in an infinite loop as neither the page nor the path gets changed.

This is one solution, I don't know if it is what you have in mind, we can also extract the page number from the "next" url.

### DIFF
--- a/apps/automation/ansible/tower/custom/api.pm
+++ b/apps/automation/ansible/tower/custom/api.pm
@@ -134,11 +134,10 @@ sub request_api {
     my ($self, %options) = @_;
 
     my $results = [];
-    my $page = 1;
+    my $path = defined($options{force_endpoint}) ? $self->{api_path} . $options{force_endpoint} : $self->{api_path} . $options{endpoint} . '?page_size=100&page=1';
 
     $self->settings();
     while (1) {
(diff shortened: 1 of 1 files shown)


garnier-quentin:
Thanks for the PR. Could you test it with last commit ?

alfredosilvestre-natixis:
Yup it is working as well, thanks

alfredosilvestre-natixis:
Hello,

In fact we just had an issue when you only have one page so the json result does not have "next" defined and it remains in an infinite loop.

Example:
{"count":4,"next":null,"previous":null,
....

garnier-quentin:
With my commit, if next is not defined, it quits the loop. Could you provide the --debug ?

alfredosilvestre-natixis:
Yes, of course, see attached.

I saw that if the next is not defined and it should work, however I think there is something on the decoded content.

[debug_output.txt](https://github.com/centreon/centreon-plugins/files/5332766/debug_output.txt)

garnier-quentin:
I have done a patch. Could you test it ?

alfredosilvestre-natixis:
Besides the extra "if" on line 170 it seems to be working 😄

garnier-quentin:
Thanks. It should be fixed

PR #2223, opened 2020-09-22
fix(win) typo

UrBnW:
Hi,

This follows https://github.com/centreon/centreon-plugins/commit/83cfe2a2465b8ffc0fd206f19243abd86e2b5c5c.

Thx 👍


UrBnW:
While here I also removed no more used configuration files.
Thank you 👍

garnier-quentin:
I prefer that syntax. I won't remove config file (it will be an example). Could you test ?

UrBnW:
Your syntax also works, perfect, thx 👍

PR #2234, opened 2020-10-02
enh(cpu) average counters

UrBnW:
Hi,

Here's a new `--average-counters` option for the standard CPU mode.
It averages the counters over the specified period of time.

Let's imagine you run this check every minute, and use `--average-counters=300`, counters will have their value set to the average of the last 5 values.

This allows to smooth peaks, and, as a result, avoid counters going above and below the warning / critical thresholds too easily / frequently.
This helps reduce number of notifications.

I wonder if this `--average-counters` could be directly part of the `counters` template, as a global option.
We would then be able to average all counters for all plugins / modes.
In this case, it would perhaps be nice to add averaged counters, instead of modifying the existing ones' value.

Thank you 👍

### DIFF
--- a/snmp_standard/mode/cpu.pm
+++ b/snmp_standard/mode/cpu.pm
@@ -19,6 +19,7 @@
 #
 
 package snmp_standard::mode::cpu;
+use centreon::plugins::statefile;
 
 use base qw(centreon::plugins::templates::counter);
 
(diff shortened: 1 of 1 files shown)


garnier-quentin:
Sorry but it looks like smarter threshold and i don't want to put a smarter thresholds system (like: threshold based on time periods).
You should have like a software doing that (like Zabbix trigger for example)

UrBnW:
Thank you for your feedback @garnier-quentin.
I thought it was a good idea for better thresholds handling / better behavior around threshold.

Should we then move to Zabbix as you seem to propose ? ^^

garnier-quentin:
They are doing some stuffs well ;) Centreon is working on anomaly detection, maybe you can take a look also:
https://docs.centreon.com/20.04/en/monitoring/anomaly-detection.html

PR #2246, opened 2020-10-12
fix multiservices

UrBnW:
Hi,

This PR fixes `multiservices` mode initially written by @Sims24.

It fixes the `--json-data` option, so that we can now properly provide inline JSON.

~It adds a `services_maxage` property so that we can be sure to look for services checked in the past X seconds (as `centreon_storage` could contain very old entries).~

Thank you 👍

### DIFF
--- a/apps/centreon/sql/mode/multiservices.pm
+++ b/apps/centreon/sql/mode/multiservices.pm
@@ -290,7 +290,7 @@ sub new {
     bless $self, $class;
 
     $options{options}->add_options(arguments => {
-        'config-file:s'     => { name => 'config_file' },
+        'config:s'          => { name => 'config' },
         'json-data:s'       => { name => 'json_data' },
         'warning-groups:s'  => { name => 'warning_groups' },
         'critical-groups:s' => { name => 'critical_groups' },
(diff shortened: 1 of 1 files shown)


Sims24:
Glad to see that some people find it useful :) Thanks for your contribution as always ;)

I'm interested to understand some of the change: 

- I can't understand why the modification is needed on the JSON parsing routine, it works for both mode on my side. This logic is used in various place without any issue. Could you explain a little bit more what doesn't work ? 
- Max_age is interesting but also dangerous as it could to a different number of Service check result included. To satisfy my curiosity, why the enabled=1 flag isn't enough for you?

UrBnW:
> Glad to see that some people find it useful :)

Useful for example to check that you have only one instance of a specific process running on a cluster 👍

> I can't understand why the modification is needed on the JSON parsing routine, it works for both mode on my side. This logic is used in various place without any issue. Could you explain a little bit more what doesn't work ?

Whatever the option you use, `--config-file` or `--json-data`, the given option-value is treated (and opened) as a file.
It's OK for `--config-file`, but not for `--json-data`, which, as far as I understand it, expects an inline JSON configuration.
The first commit of this PR then fixes this.

> Max_age is interesting but also dangerous as it could to a different number of Service check result included. To satisfy my curiosity, why the enabled=1 flag isn't enough for you?

Let's imagine you are checking `hostA/serviceA`.
It works, mode returns 1 matching service.
Now you (or someone else, accidentally or not, whatever) delete `serviceA`.
Mode will still return 1 matching service, as the old entry remains in `centreon_storage` database. First issue then.
You recreate `serviceA`.
Mode will now return 2 matching services :|

Checking `services.last_check` allows then to be sure to check the proper running services.
You have to set it somewhere around your `check_interval` to be safe, as it's more a security measure than a proper filter.
This is the second commit of the PR.

Thx 👍

Sims24:
Ok, for the JSON config this is not what I expected, I probably never fixed it. 

@garnier-quentin implemented a much more elegant way to achieve that here: https://github.com/centreon/centreon-plugins/blob/eee2b808cdfbe52f466d12e7c924435e9c48241d/apps/protocols/modbus/mode/numericvalue.pm#L184. Could you please stick to it, it will avoid having two options where one is enough.

Regarding the flag, it should never stay enabled if the service is deleted and the engine reloaded nor restarted ... So if it returns two Service, there is an issue somewhere 🤔

UrBnW:
> Could you please stick to it, it will avoid having two options where one is enough.

Done 👍

> Regarding the flag, it should never stay enabled if the service is deleted and the engine reloaded nor restarted ... So if it returns two Service, there is an issue somewhere 🤔

Strangely enough, I was not able to reproduce this issue again...
Rather weird, I don't understand...
I've then removed the `services_maxage` property for now.

Thank you again for your review 👍

Sims24:
Thank you !

Sims24 (approved):

UrBnW:
Thank you @Sims24 for the merge 👍

Last question for both of U @Sims24 @garnier-quentin, how do you run `--plugin=database::mysql::plugin --dyn-mode=apps::centreon::sql::mode::multiservices` using Perl Connector, as `apps::centreon::sql::mode::multiservices` is not embedded into `database::mysql::plugin` FatPacked plugin ?
Many thx 👍

UrBnW:
OK, found it, simply using `centreon-plugin-Applications-Monitoring-Centreon-Database`.

PR #2254, opened 2020-10-14
Update vplexapi.pm error line 137

Le-Syl21:
Correct an error:
/storage/emc/vplex/restapi/custom/vplexapi.pm line 137
when clusters as been renamed (not named by default cluster-1 & cluster-2)
Bug exist on VS2 & VS6 patch tried on both and works greats
All mode commandes on both VS2 & VS6 in debug mode:
VS2 -> http://dl.free.fr/fXGIMpctN
VG6 -> http://dl.free.fr/sWvDo8tYA


garnier-quentin:
Thanks. Could you test with last version ? I have done more changes. I enhanced filter options also

PR #2341, opened 2020-11-12
enh(mode): alyvix testcase date & freshness

thibaults-centreon:


garnier-quentin:
Is 'testcase-date' useful ? We can display the last freshness in date (not seconds) for the output

thibaults-centreon:
This counter aims to get this long output:
checking test case 'xxxxx'
    duration: 13390 ms, state: FAILED, last execution: 2020-11-12T14:21:18 (15s ago)
whereas the thresholds will only apply to the freshness value, so the "not ok" short output is more relevant in seconds.

Is there a more efficient way to do it ?

garnier-quentin:
Like that:
```
{ label => 'testcase-freshness', display_ok => 0, set => {
                key_values => [ { name => 'freshness' }, { name => 'lastexec' },  { name => 'display' } ],
                closure_custom_output => $self->can('custom_date_output'),
                closure_custom_perfdata => sub { return 0; },
            }
        }
}
```

garnier-quentin:
Could you add the perfdata ? It could be useful.

PR #2383, opened 2020-11-27
enh(mikrotik) Add firmware mode

UrBnW:
Hi,

This PR adds a `firmware` mode to MikroTik plugin.
So that we can check that both software and firmware run the same version (which is expected, but could not be the case as both have their separate update process).

`OK: 'RouterOS CRS326-24S+2Q+' : software is '6.47.7', firmware is '6.47.7'`

Thx 👍


UrBnW:
Output example :
`OK: RouterOS CRS326-24S+2Q+ : software is '6.47.7', firmware is '6.47.7'`

PR #2393, opened 2020-12-01
Breakdown of branches and update of OIDs for switches using HiOS OS

LLFT:
I only updated the OIDs and deleted the hardware function which is no longer usable as it is on HiOS.
I am not affected to Standard, I duplicated it to rename it L2P

== 

Je n'ai mis à jour que les OID et supprimé la fonction hardware qui n'est plus exploitable en l'état sur les HiOS. 
Je n'est pas touché à Standard, je l'ai dupliqué pour le renommer L2P


garnier-quentin:
I need a full snmpwalk and mibs files. I cannot change it like that. Need to think how to do it (maybe manage old and new).

LLFT:
I understand ! Then it's more reasonable, which is why I left the Standard. I wasn't asking for the merge anyway. But only show you the changes made.

Attached are the requested files.
[Hirschmann.zip](https://github.com/centreon/centreon-plugins/files/5622369/Hirschmann.zip)

LLFT:
I intended to add backup control to Hirschmann controls.
And the control of the CRC and Collision thresholds in the snmp_standard
CRC_MIB = ". 1.3.6.1.2.1.16.1.1.1.8"
COLLISION_MIB = ". 1.3.6.1.2.1.16.1.1.1.13"

But suddenly I submit the idea to you.

garnier-quentin:
You can test it with the same plugin your 2 devices. It should work. I have renamed 'backup' mode by 'configuration'.

PR #2411, opened 2020-12-04
fix(win) make --list-plugin Windows compatible

### LINKED ISSUE #2409: --list-plugin does not work on Windows
Hi,

Strangely enough, `--list-plugin` option does not work on Windows (with `centreon_plugins.exe`), it returns nothing at all.

Thx 👍

UrBnW:
Hi,

This makes `--list-plugin` Windows compatible 👍 
It then fixes #2409.

Thx !


garnier-quentin:
You can test. I prefer that method. Par packer can be also used for Linux

PR #2451, opened 2020-12-22
bug fix when queue has no message

matoy:
just add a check when object is undefined.
this is to avoid error message: UNKNOWN: Can't use an undefined value as an ARRAY reference at /usr/lib/centreon/plugins/centreon-plugins/apps/mulesoft/restapi/mode/messages.pm line 150
and the current check to fail.


garnier-quentin:
Thanks for the PR!

PR #2524, opened 2021-01-18
enh(plugin)apps-protocol-http metricsv2

itoussies:


apps/protocols/http/plugin.pm
```diff
@@ -30,12 +30,12 @@ sub new {
     bless $self, $class;
 
     $self->{version} = '0.1';
-    %{$self->{modes}} = (
-        'expected-content'  => 'apps::protocols::http::mode::expectedcontent',
-        'json-content'      => 'apps::protocols::http::mode::jsoncontent',
-        'response'          => 'apps::protocols::http::mode::response',
-        'soap-content'      => 'apps::protocols::http::mode::soapcontent',
-    );
+    $self->{modes} = (
```
Sims24, line 33: ( instead of {

Sims24 (changes requested):

Issue #2505, opened 2021-01-13
[os::windows::local::plugin] Mode sessions : filter-sessionname option


mickael-decastro:
Hi, 

With the current centreon-plugin version (Global Version: 20210107 ) , the filter-sessionname option doesn't work property.

Here a command check number of sessions rdp.
```
PS C:\Program Files\Centreon NSClient++\scripts\centreon> .\centreon_plugins.exe --plugin "os::windows::local::plugin" --mode "sessions" --language fr  --filter-sessionname ".*rdp*" --debug
**OK: Sessions created: 0, disconnected: 0, reconnected : 0, current active : 0, current disconnected : 0 | 'sessions_created'=0;;;0; 'sessions_disconnected'=0;;;0; 'sessions_reconnected'=0;;;0; 'sessions_active'=0;;;0; 'sessions_disconnected_current'=0;;;0;
 SESSION           UTILISATEUR              ID  ÉTAT    TYPE        PÉRIPHÉRIQUE
 services                                    0  Déco
 console                                     1  Conn
>rdp-tcp#10        user-test        12  Actif
 ica-cgp                                 65536  Écouter
 ica-html5                               65537  Écouter
 ica-tcp                                 65538  Écouter
 rdp-tcp                                 65539  Écouter
Nombre total de sessions créées : 13
Nombre total de sessions déconnectées : 10
Nombre total de sessions reconnectées : 0
skipping 'services': no matching filter.
skipping 'console': no matching filter.
**skipping '-': no matching filter.
skipping 'ica-cgp': no matching filter.
skipping 'ica-html5': no matching filter.
skipping 'ica-tcp': no matching filter.
```
The session ">rdp-tcp#10" is skip with "-" in stead of the right session name

Here the output when I set --filter-sessionname with value "-". We can see the rdp user session is properly detected.
```
PS C:\Program Files\Centreon NSClient++\scripts\centreon> .\centreon_plugins.exe --plugin "os::windows::local::plugin" --mode "sessions" --language fr  --filter-sessionname "-" --debug
**OK: Sessions created: 0, disconnected: 0, reconnected : 0, current active : 1, current disconnected : 0 | 'sessions_created'=0;;;0; 'sessions_disconnected'=0;;;0; 'sessions_reconnected'=0;;;0; 'sessions_active'=1;;;0; 'sessions_disconnected_current'=0;;;0;
[...]

PR #2528, opened 2021-01-19
fix(win) Sessions mode run manually

UrBnW:
Hi,

This should solve #2505 (not tested though).

Thx 👍


garnier-quentin:
it's not enough

PR #2532, opened 2021-01-20
fix(mode): Fix Velocloud modes when edge offline

thibaults-centreon:
This avoids the  "OK:  | " output for the modes when the edge is offline or unavailable


garnier-quentin:
I think it won't work. For example: the mode will exit if you have one edge without applications. I prefer to add a counter: ```edge.applications.total.count```. With that counter, the user can add a threshold if a edge doesn't have applications.

thibaults-centreon:
I don't understand. If an edge doesn't have any applications or categories, then the related modes shouldn't be used for the edge. What's the point to monitor something that doesn't exist ?

garnier-quentin:
You can have X edges with that plugin. If one has no application, maybe we want to check other also (and get perfdatas).

garnier-quentin (approved):

PR #2571, opened 2021-02-05
fix for a case where the lookup cannot find anything

Coolgeek789:
This is a fix for the case when the response of the soap request doesn't contain what we're looking for (the lookup doesn't match anything)


garnier-quentin:
You can use that option: ```--critical-numeric=1:```

Coolgeek789:
yup that works too....
sorry for the noise !

PR #2572, opened 2021-02-05
enh(plugin)network-checkpoint-snmp: add new gateway mode

itoussies:
New mode to check the default gateway.


garnier-quentin:
It's not really "monitoring". I won't merge it.

PR #2660, opened 2021-03-17
add(plugin): Azure FrontDoor

thibaults-centreon:


cloud/azure/custom/api.pm
```diff
@@ -334,7 +334,7 @@ sub azure_get_resource_health_set_url {
 
     my $url = $self->{management_endpoint} . "/subscriptions/" . $self->{subscription} . "/resourceGroups/" .
         $options{resource_group} . "/providers/" . $options{resource_namespace} . "/" . $options{resource_type} .
-        "/" . $options{resource} . "/providers/Microsoft.ResourceHealth/availabilityStatuses/current?api-version=" . $self->{api_version};
+        "/" . $options{resource} . "/providers/Microsoft.ResourceHealth/availabilityStatuses/current?api-version=" . $options{api_version};
```
Sims24, line 337: @thibaults-centreon , could you explain why this is needed? I'm probably missing something.

thibaults-centreon: the ResourceHealth endpoint needs a specific API version which is hardcoded here: https://github.com/centreon/centreon-plugins/blob/master/cloud/azure/management/monitor/mode/health.pm#L116
Then the value has to be fetched from the options.

Sims24: Heyhey of course. Thanks !

Sims24 (approved):

PR #2675, opened 2021-03-25
enh(plugin) add metricnamespace option to Azure Monitor get-metric mode

### LINKED ISSUE #2248: enh(plugin) Azure Monitor Metric Namespace management
When it comes to custom Azure monitor metric, it's mandatory to specify the metric_namespace.

We have to add it in the get-metric mode. 

Thx

petneli:
Added metricnamespace as parameter to the Azure Metrics script

Fixes #2248

### DIFF
--- a/cloud/azure/management/monitor/mode/getmetrics.pm
+++ b/cloud/azure/management/monitor/mode/getmetrics.pm
@@ -99,7 +99,8 @@ sub new {
         "resource-type:s"       => { name => 'resource_type' },
         "resource-namespace:s"  => { name => 'resource_namespace' },
         "metric:s@"             => { name => 'metric' },
-        "filter-dimension:s"    => { name => 'filter_dimension'}
+        "filter-dimension:s"    => { name => 'filter_dimension'},
+        "metricnamespace:s"     => { name => 'metricnamespace' }
     });
 
     return $self;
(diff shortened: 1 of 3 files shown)


Sims24 (commented):
Hi @petneli 

Thanks for your contribution. It's cool to have a new contributor, welcome :)

Few questions/remarks: 

- Could you share what do you plan to do with the 'Count' aggregation? I'm interested in the usecase behind. We didn't add it as it returns the number of records but doesn't deal with the measurement itself.
- Could you also update the azcli.pm custommode? We try to have same feature in both.

Simon

cloud/azure/management/monitor/mode/getmetrics.pm
```diff
@@ -230,6 +235,10 @@ Set resource type (Required if resource's name is used).
 
 Set monitor metrics (Required) (Can be multiple).
 
+=item B<--metricnamespace>
```
Sims24, line 238: Tiny detail, could you name it --metric-namespace with a dash as word separator ;)

petneli:
Hi @Sims24,

- We have defined custom metric with aggregation "count" and couldn't monitor it + it is in general supported by Azure, so I thought it is better to have it then not. Interesting use cases would be e.g. to see the number of data points in case of missing data (see "More complex example" [metrics-aggregation-explained](https://docs.microsoft.com/en-us/azure/azure-monitor/essentials/metrics-aggregation-explained) )
- I have adjusted azcli.pm as well. ( added "count" as aggregation and "metric namespace" parameter)
- I have renamed the parameter "--metricnamespace" to "--metric-namespace" as to comply with the general style.

Neli

Sims24:
Wonderful!

Thank you

Sims24 (approved):

PR #2692, opened 2021-04-01
enh(security): Add SELinux rules

lpinsivy:
Add SElinux rules


garnier-quentin (approved):

garnier-quentin (commented):
please set it in contrib directory.

PR #2709, opened 2021-04-15
(plugin) apps::protocols::tcp - add responsetime mode and rename existing one into connectionstatus

UrBnW:
Hi,

This PR adds a proper TCP `responsetime` mode which mimics :
- `check_icmp` binary ;
- `apps::protocols::snmp::mode::responsetime` mode.

The main purpose of this new TCP `responsetime` mode is to be used as a host check, when `check_icmp` can't be used.
Think about for example of a Windows host (with NSClient++) behind a router / firewall.
This new TCP `responsetime` mode will then allow to check the availability of host, computing proper already known `rta` / `rtmin` / `rtmax` / `pl` perfdata.

Current `apps::protocols::tcp::mode::responsetime` mode has then been renamed to a more appropriated `apps::protocols::tcp::mode::connectionstatus` mode.

Many thx 👍

### DIFF
--- a/apps/protocols/tcp/mode/responsetime.pm
+++ b/apps/protocols/tcp/mode/responsetime.pm
@@ -1,0 +1,181 @@
+#
+# Copyright 2021 Centreon (http://www.centreon.com/)
+#
+# Centreon is a full-fledged industry-strength solution that meets
+# the needs in IT infrastructure and application monitoring for
+# service performance.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+package apps::protocols::tcp::mode::responsetime;
+
+use base qw(centreon::plugins::templates::counter);
+
+use strict;
+use warnings;
+use Time::HiRes qw(gettimeofday tv_interval);
+use IO::Socket::INET;
+
+sub prefix_output {
+    my ($self, %options) = @_;
+
+    return 'TCP port ' . $self->{option_results}->{port} . ' ';
+}
+
+sub set_counters {
+    my ($self, %options) = @_;
+    
+    $self->{maps_counters_type} = [
+        { name => 'global', type => 0, cb_prefix_output => 'prefix_output' }
+    ];
+
+    $self->{maps_counters}->{global} = [
+        { label => 'rta', nlabel => 'roundtrip.time.average.milliseconds', set => {
+                key_values => [ { name => 'rta' } ],
+                output_template => 'rta %.3fms',
+                perfdatas => [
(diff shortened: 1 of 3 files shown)


UrBnW:
```
# centreon_plugins.pl --plugin=apps::protocols::tcp::plugin --mode=response-time --hostname=google.com --port=80  
OK: TCP port 80 rta 9.529ms, lost 0% | 'rta'=9.529ms;;;0; 'rtmax'=10.925ms;;;0; 'rtmin'=8.965ms;;;0; 'pl'=0%;;;0;100
```

apps/protocols/tcp/mode/responsetime.pm
```diff
@@ -99,105 +93,98 @@ sub check_options {
...
+    } else {
         $self->{output}->add_option_msg(short_msg => 'Please set the port option');
         $self->{output}->option_exit();
     }
+    if (defined($self->{option_results}->{timeout}) && $self->{option_results}->{timeout} =~ /(\d+)/) {
+        $self->{option_timeout} = $1;
+    }
+    if (defined($self->{option_results}->{packets}) && $self->{option_results}->{packets} =~ /(\d+)/) {
+        $self->{option_packets} = $1;
+    }
```
Sims24, line 109: You should use the following syntax to define a default value for options: 

```
        'timeout:s'  => { name => 'timeout' , default => 5},
        'packets:s'  => { name => 'packets' , default => 5},
```

apps/protocols/tcp/mode/responsetime.pm
```diff
@@ -99,105 +93,98 @@ sub check_options {
...
-    }
-    if (!defined($self->{option_results}->{port})) {
+    $self->{option_timeout} = 5;
+    $self->{option_packets} = 5;
+    if (defined($self->{option_results}->{port}) && $self->{option_results}->{port} =~ /(\d+)/) {
+        $self->{option_port} = $1;
+    } else {
         $self->{output}->add_option_msg(short_msg => 'Please set the port option');
         $self->{output}->option_exit();
     }
```
Sims24, line 103: You can simply check that an option is defined and not empty. You can of course check that's an integer but we don't got this far most of the time ...

apps/protocols/tcp/mode/responsetime.pm
```diff
@@ -99,105 +93,98 @@ sub check_options {
...
+    my $total_time_elapsed = 0;
+    my $max_time_elapsed = 0;
+    my $min_time_elapsed = 0;
+    my $total_packet_lost = 0;
+    for (my $i = 0; $i < $self->{option_packets}; $i++) {
+        my $timing0 = [gettimeofday];
+        my $return = IO::Socket::INET->new(
             PeerAddr => $self->{option_results}->{hostname},
-            PeerPort => $self->{option_results}->{port},
-            Timeout => $self->{option_results}->{timeout},
```
Sims24, line 127: Better to keep the common way of accessing value option IMHO

Sims24 (changes requested):
Few things that need to be done. 

I agree it's a good idea to have a much detailed tcp ping, nevertheless renaming modes require some changes on the Plugin Pack packaging and it will take some time to merge it and avoid breaking existing setup.

UrBnW:
Many thanks for your review @Sims24, all points addressed in last commit !

Sims24 (approved):

UrBnW:
Many thanks @Sims24 👍

UrBnW:
Thank you very much @garnier-quentin ;)

PR #2746, opened 2021-04-29
enh(Synology) Use disk name rather than its ID

UrBnW:
Hi,

Let's use Synology disk name, rather than its SNMP ID, for better handling.

```
Checking disk
disk 'Drive 1' status is Normal [instance: 0]
disk 'Drive 2' status is Normal [instance: 1]
disk 'Drive 3' status is Normal [instance: 2]
disk 'Drive 4' status is Normal [instance: 3]
disk 'Drive 5' status is Normal [instance: 4]
disk 'Drive 6' status is Normal [instance: 5]
disk 'Drive 7' status is Normal [instance: 6]
disk 'Drive 8' status is Normal [instance: 7]
disk 'Drive 9' status is Normal [instance: 8]
disk 'Drive 10' status is Normal [instance: 9]
disk 'Drive 11' status is Initialized [instance: 10]
disk 'Drive 12' status is Initialized [instance: 11]
```

Thx 👍

### DIFF
--- a/storage/synology/snmp/mode/components/disk.pm
+++ b/storage/synology/snmp/mode/components/disk.pm
@@ -32,13 +32,19 @@ my $map_disk_status = {
 };
 
 my $mapping = {
+    synoDiskdiskName   => { oid => '.1.3.6.1.4.1.6574.2.1.1.2' },
     synoDiskdiskStatus => { oid => '.1.3.6.1.4.1.6574.2.1.1.5', map => $map_disk_status }
 };
+my $oid_synoDisk = '.1.3.6.1.4.1.6574.2.1.1';
 
 sub load {
     my ($self) = @_;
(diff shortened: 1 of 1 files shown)


Sims24 (approved):

Sims24:
@garnier-quentin I merged this one but maybe it wasn't a good idea. Will it break anything in the user land (filter in checks?)

UrBnW:
Many thx for the merge @Sims24 👍

> Will it break anything in the user land (filter in checks?)

Will not, here we changed only the output name, filter is on instance number.
In this regard, filter on name would perhaps be more convenient / more relevant than on SNMP instance number (as per the output in the first post), but then it's the case for many / most of the plugins, and would here be a breaking change.
Thx again !

PR #2751, opened 2021-04-29
enh(equallogic) Add some standard modes

UrBnW:
Hi,

Let's add time and uptime modes to equallogic plugin.

Thx 👍


garnier-quentin:
We can inherit the mode standard and overload.

UrBnW:
Do you want me to perform such modification ?

garnier-quentin:
I will do it ;)

PR #2757, opened 2021-04-30
enh(equallogic) Add model and version in health status

UrBnW:
Hi,

As already done for some other hardwares, let's output model and version number in heath status.

Thx 👍

```
# ./centreon_plugins.pl --plugin=storage::dell::equallogic::snmp::plugin --mode=hardware ... --verbose --component=health
OK: All 1 components are ok [1/1 health]. | 'count_health'=1;;;;
Checking health
PS4110.10.0.3 'EQL1-BACKUP' health status is normal [instance: 1.1606743924].
```

### DIFF
--- a/storage/dell/equallogic/snmp/mode/components/health.pm
+++ b/storage/dell/equallogic/snmp/mode/components/health.pm
@@ -31,15 +31,33 @@ my %map_health_status = (
 );
 
 # In MIB 'eqlcontroller.mib'
-my $mapping = {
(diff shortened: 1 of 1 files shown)


UrBnW:
While here, one question, how to have `PS4110.10.0.3 'EQL1-BACKUP' health status is normal` in short output directly (when there's only one component checked, as in the above example) ?
Would be convenient.
Thx 👍

PR #2821, opened 2021-05-24
enh(plugin): SAHI Pro REST API metricsv2

itoussies:
Updating the plugin with the new metric naming format.

### DIFF
--- a/apps/sahipro/restapi/mode/scenario.pm
+++ b/apps/sahipro/restapi/mode/scenario.pm
@@ -24,7 +24,7 @@ use base qw(centreon::plugins::templates::counter);
 
 use strict;
 use warnings;
-use centreon::plugins::templates::catalog_functions qw(catalog_status_threshold);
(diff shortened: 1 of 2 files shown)


apps/sahipro/restapi/mode/scenario.pm
```diff
@@ -62,99 +74,89 @@ sub set_counters {
...
                 output_template => 'execution time : %s ms',
                 perfdatas => [
-                    { label => 'total_time', value => 'time_taken', template => '%s', min => 0, unit => 'ms' },
-                ],
+                    { label => 'total_time', template => '%s', min => 0, unit => 'ms' }
+                ]
             }
         },
-        { label => 'total-steps', display_ok => 0, set => {
+        { label => 'total-steps', nlabel => 'scenario.steps.total.count', set => {
```
Sims24, line 98: Don't use total, count is enough.

Sims24 (changes requested):

Sims24 (approved):

PR #2834, opened 2021-05-27
reorg(plugin)make nmap a dedicated plugin

Sims24:


apps/nmap/cli/mode/discoverynmap.pm
```diff
@@ -18,13 +18,12 @@
 # limitations under the License.
 #
 
-package os::linux::local::mode::discoverynmap;
+package apps::nmap::cli::mode::discoverynmap;
```
thibaults-centreon, line 21: Shouldn't it be just "discovery" with the mode being named "discovery.pm" ? What do you think ?

Sims24: You're obviously right!
Done.

thibaults-centreon (approved):

PR #2905, opened 2021-06-24
fix(synology) properly rename temperature metric

UrBnW:
Hi,

Let's properly rename Synology temperature metric as in other hardware modes.

Thx 👍

### DIFF
--- a/storage/synology/snmp/mode/temperature.pm
+++ b/storage/synology/snmp/mode/temperature.pm
@@ -33,7 +33,7 @@ sub set_counters {
     ];
 
     $self->{maps_counters}->{global} = [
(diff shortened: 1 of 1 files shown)


garnier-quentin:
Why do you want to change the name ? And that not the right way to add the instance.

UrBnW:
Because temperature metric is already named `hardware.temperature.celsius` in many various plugins, so it's to make things consistent.
Main reason is that it's easier to template once metrics are in time-series databases, querying metrics with the same name...
Thank you 👍

garnier-quentin:
Fixed: https://github.com/centreon/centreon-plugins/pull/2915

PR #2910, opened 2021-06-29
Wrong command powershell command line link with case 00097542

Tpo76:
Wrong command PowerShell command-line link with case 00097542


garnier-quentin:
@Tpo76 Could you test directly with:
```
-PSProvider CMSite
```

Tpo76:
The PSProvider option doesn't work even with PowerShell run as admin.
The plugin output is UNKNOWN: No sites found.

There is sites on the server which are detected with the command provided by this commit.

PR #2916, opened 2021-06-30
fix(mulesoft): http content type error

garnier-quentin:


thibaults-centreon:
I'm a bit concerned that this would break things on some platforms. 
The "application/x-www-form-urlencoded" content type was set on purpose because it was the only way to make the API work on the test environment I used for the development.
Maybe they changed it on the latest version of the API but sadly it is not documented on Mulesoft site :(

garnier-quentin:
There is nothing in the POST (it's a get request). It shouldn't break. If someone have the issue, we'll try to fix it.

PR #2922, opened 2021-07-01
add(n1100) Add Dell N1100 series plugin

UrBnW:
Hi,

Let's add a new plugin, for Dell N1100 series network switches.

Many thanks :+1:


UrBnW:
Would be great if you could also create the `rpm` package for this new plugin.
Thank U again !

garnier-quentin:
I don't want to continue to create x version for dell. Need to have something about dell os type (radlan,marvel,...)

PR #2968, opened 2021-07-16
New plugin ACR (Azure Container Registry)

matoy:
New plugin ACR (Azure Container Registry)

### DIFF
--- a/cloud/azure/storage/acr/mode/storageused.pm
+++ b/cloud/azure/storage/acr/mode/storageused.pm
@@ -1,0 +1,143 @@
+#
+# Copyright 2021 Centreon (http://www.centreon.com/)
+#
+# Centreon is a full-fledged industry-strength solution that meets
+# the needs in IT infrastructure and application monitoring for
+# service performance.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


Sims24:
Hello @matoy 

Thanks for this contribution (and the past ones), it means and helps a lot! 

I checked the supported metrics for ACR and I only found these: https://docs.microsoft.com/fr-fr/azure/azure-monitor/essentials/metrics-supported#microsoftcontainerregistryregistries. 

Maybe I'm not looking at the right place?

Thanks

matoy:
indeed, it is not stated in the MS doc; but i can assure you the metric exists...
and the check works well in our centreon instance.

Capture d’écran 2021-07-16 145058

Sims24:
Ok thanks for confirming that MS Doc is not up-to-date --'. 

We will merge it and maybe add few modes to get the percentage of successful run. WOuld it make sense? 

Regards

matoy:
to be honest, i did write other modes because we didn't find any purpose to them but that might interest other people :)

Sims24:
Ok, you're the user so if it doesn't make sense we will wait for somebody to ask for it. 

Thanks

Sims24 (approved):

PR #2978, opened 2021-07-21
enh(synology) Some useful comment from support

UrBnW:
Hi,

Some useful comment from Synology support which may help others (like me) wondering about this OID behavior.

Thx 👍


Sims24 (approved):

Sims24:
Thanks for sharing this with the community 👍

Issue #1651, opened 2019-09-23
Negative value disk : skipping storage 'E:': total size is <= 0 (-1591421702144)


kevin-m87:
Hello, i'm facing this problem of negative value on my HDD with 15To :
skipping storage 'E:': total size is <= 0 (-1591421702144)

I know its a problem with integer32.

I'm using centreon_windows_snmp.pl we can add some some code for resolve this or do something ?

Thanks for your help.

Sims24:
Hello Kevin, 

Thanks for your message but infortunately no we do not have any solution here ..

Issue #983, opened 2018-05-09
os::windows::snmp::plugin is giving me negative value


MikeW83:
Hello community,

this is my command:
```
/usr/lib/centreon/plugins/centreon_windows_snmp.pl
	--plugin=os::windows::snmp::plugin
	--mode=storage
	--hostname=xxx.xxx.xxx.xxx
	--snmp-version=2c
	--snmp-community=public 
	--storage '2'
	--display-transform-src='^(..).*'
	--display-transform-dst='$1'
	--warning-usage='80'
	--critical-usage='90'
```

And this is the result:

> skipping storage 'D:': total size is <= 0 (-15183382585344)

How can I fix this? It only happens with one (!) service. Also deleting and recreating the service doesn't resolve this issue.

Thank you!

MikeW83:
I have done a snmpwalk with this result:

> HOST-RESOURCES-MIB::hrMemorySize.0 = INTEGER: 16679272 KBytes
> HOST-RESOURCES-MIB::hrStorageIndex.1 = INTEGER: 1
> HOST-RESOURCES-MIB::hrStorageIndex.2 = INTEGER: 2
> HOST-RESOURCES-MIB::hrStorageIndex.3 = INTEGER: 3
> HOST-RESOURCES-MIB::hrStorageIndex.4 = INTEGER: 4
> HOST-RESOURCES-MIB::hrStorageIndex.5 = INTEGER: 5
> HOST-RESOURCES-MIB::hrStorageType.1 = OID: HOST-RESOURCES-TYPES::hrStorageFixedDisk
> HOST-RESOURCES-MIB::hrStorageType.2 = OID: HOST-RESOURCES-TYPES::hrStorageFixedDisk
> HOST-RESOURCES-MIB::hrStorageType.3 = OID: HOST-RESOURCES-TYPES::hrStorageCompactDisc
> HOST-RESOURCES-MIB::hrStorageType.4 = OID: HOST-RESOURCES-TYPES::hrStorageVirtualMemory
> HOST-RESOURCES-MIB::hrStorageType.5 = OID: HOST-RESOURCES-TYPES::hrStorageRam
> HOST-RESOURCES-MIB::hrStorageDescr.1 = STRING: C:\\ Label:OS
> HOST-RESOURCES-MIB::hrStorageDescr.2 = STRING: D:\\ Label:DATA
> HOST-RESOURCES-MIB::hrStorageDescr.3 = STRING: E:\\
[...]

PR #3043, opened 2021-08-12
Updated storage.pm to work with larger volumes

BlackCertz:
Hey, I've been told in Slack to suggest this change over here. So please be gentle as this is my first GitHub post 😬.

In reference to:
https://github.com/centreon/centreon-plugins/issues/1651
https://github.com/centreon/centreon-plugins/issues/983

Since the Windows SNMP agent utilizes signed 32-bit Integer variables (-2.147.483.648 to 2.147.483.647) for storing the size and used amounts in bytes of disks, they are prone to overflow and report negative values when working with large disks of 8TB+. 

This simple workaround checks for a negative return value and adds the 32-bit range (2^32 = 4.294.967.296).
From personal experience, this works with disks of up to 40TB.
I am however not sure of the actual limit this works with.
Furthermore, with a 40TB disk, 4096 Bytes block size (= 10.742.787.580 hrStorageSize), it should wrap around multiple times, however the plugin still manages to report the correct value with this change (As to why... I'd be very interested to hear if you could tell me! 😄 ).

### DIFF
--- a/snmp_standard/mode/storage.pm
+++ b/snmp_standard/mode/storage.pm
@@ -335,8 +335,18 @@ sub manage_selection {
             next;
         }
         
+        my $size = $result->{$oid_hrStorageSize . "." . $_};
+        my $used = $result->{$oid_hrStorageUsed . "." . $_};
+        
+        if ($size <= 0) {
+            $size += 2**32;
+        }
+        if ($used <= 0) {
+            $used += 2**32;
+        }
+        
         # in bytes hrStorageAllocationUnits
-        my $total_size = $result->{$oid_hrStorageSize . "." . $_} * $result->{$oid_hrStorageAllocationUnits . "." . $_};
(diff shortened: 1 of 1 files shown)


garnier-quentin:
Thanks for the PR. It's working in some cases but it stil quite precarious. Maybe i prefer to add an option if it's negative (than automatic).

BlackCertz:
Yup, it's a rather simple workaround. 
Although here's some practical examples of storage volumes I'm currently monitoring with this fix:

This drive should report a positive value (One wrap | Positive -> Negative -> Positive):
hc_751
This one should report a negative value (One wrap | P -> N):
hc_752
And this one should also report a negative value (Three wraps | P -> N -> P -> N -> P -> N):
hc_753

garnier-quentin:
You can test with last git version with option --counters-overflow

PR #3045, opened 2021-08-16
(plugin) apps::proxmox::ve::restapi - add filtering capabilities to storageusage.pm

OlivierKuziemko:
Add filtering by node ID and node name.


garnier-quentin:
i don't understand how it can work. The method ```api_get_storages``` doesn't use params ```node_name``` and ```node_id```

PR #3046, opened 2021-08-16
(plugin) apps::protocols::nrpe - enhance perfdata management in NRPE custom mode

OlivierKuziemko:
Updating nrpe protocol to make it works with filter-perfdata while using custom mode nrpe.

### DIFF
--- a/apps/protocols/nrpe/custom/nrpe.pm
+++ b/apps/protocols/nrpe/custom/nrpe.pm
@@ -121,7 +121,7 @@ sub output_noperf {
 sub format_result {
     my ($self, %options) = @_;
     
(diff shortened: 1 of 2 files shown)


garnier-quentin:
I'm not sure it's very useful. If you use nsclient, you can filter perfdata with nsclient.
And if you use nrpe on linux, you use centreon-plugins obviously so you can filter perfdata ;)

Tell me the case. Maybe we could implement it with an option (not by default).

garnier-quentin:
Ok i see. thanks for the feedback

Sims24:
@OlivierKuziemko 

I will check with Quentin. In the meantime, could you remove the proxmox file that is erroneously included in this PR please?

garnier-quentin:
https://github.com/centreon/centreon-plugins/pull/3703

You have the option ```--parse```

PR #3052, opened 2021-08-20
add memory monitoring for stormshield appliance

dmagniez:
Actual memory mode is wired to memory from OS side of appliance, which is not ideal, as memory is mainly cached and full most of time.
Stormshield give access to memory usage from the software side.
This new mode, allow to monitor stormshield memory usage from the internal software side.


garnier-quentin:
Thanks. could you provide the snmpwalk and stormshield mibs files ?

garnier-quentin:
Thanks for the PR. You can test with the last git version the new mode ```memory-detailed```. I don't want to rename the memory in os.

PR #3077, opened 2021-09-03
add(plugin): Gude UPC8226 SNMP

thibaults-centreon:


garnier-quentin (commented):
Could you manage more hardware version ?
I think we can do a single plugin for "Switched & Metered Power Distribution Units" versions. And we can manage the 2 kinds of products:
- Outlet-Metered: you can get metrics for each port of the X load outputs
- Unit-Metered

garnier-quentin:
Otherwise could you get gude snmpwalk with sysdesc OID ?

garnier-quentin:
The plugin: https://github.com/centreon/centreon-plugins/pull/3268

PR #3227, opened 2021-11-02
Add size in liststorages disco format

cedricmeschin:
Add size in liststorages disco format


os/linux/local/mode/liststorages.pm
```diff
@@ -112,6 +112,7 @@ sub disco_show {
             name => $name,
             fs => $results->{$name}->{fs},
             type => $results->{$name}->{type},
+	    size => $results->{$name}->{size},
```
Sims24, line 115: @cedricmeschin please indent correctly here ;)

PR #3274, opened 2021-11-24
plugin(add) apps-dynamics-ax

Tpo76:
init apps-dynamics-ax plugin
check EDI order error from database


Sims24 (commented):
Thanks for this, good start. 

I think it would be possible to save some code by applying the logic Colin used with Toshiba Storemate (https://github.com/centreon/centreon-plugins/tree/2fd7bbb451f3dbf44b193d3566649bd0c2ed8c3e/apps/toshiba) .

This way you can call the mode like this: --plugin=database::mssql::plugin --dyn-mode=apps::dynamics::ax::mode::ediorder

Also, could you change the perfdata name and delete the 'total', it' s not necessary as total might be ambiguous (by convention it relates to a counter type), just count is enough.

garnier-quentin (approved):

Issue #3277, opened 2021-11-25
[apps::backup::veeam::local::mode::tapejobs] Use of unitialized value in hash element [..] line 171


Sims24:
Hi, 

We should handle here the case where we get a valid JSON with null values because it triggers initialized values: 

- https://github.com/centreon/centreon-plugins/blob/5f6505004c2f122f2957a4531585938df9ed8c0f/apps/backup/veeam/local/mode/tapejobs.pm#L164

Example of problematic JSON: 

```
[{"lastResult":null,"name":null,"lastState":null,"type":null,"enabled":null}]
```

PR #3280, opened 2021-11-26
(plugin) patch edge case on empty JSON with Veeam tape-jobs mode

Sims24:
Fix https://github.com/centreon/centreon-plugins/issues/3277

### DIFF
--- a/apps/backup/veeam/local/mode/tapejobs.pm
+++ b/apps/backup/veeam/local/mode/tapejobs.pm
@@ -161,6 +161,11 @@ sub manage_selection {
             $self->{output}->output_add(long_msg => "skipping job '" . $job->{name} . "': no matching filter type.", debug => 1);
             next;
         }
+        # Sometimes we may get such JSON: [{"lastResult":null,"name":null,"lastState":null,"type":null,"enabled":null}]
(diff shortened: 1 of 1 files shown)


apps/backup/veeam/local/mode/tapejobs.pm
```diff
@@ -161,6 +161,11 @@ sub manage_selection {
             $self->{output}->output_add(long_msg => "skipping job '" . $job->{name} . "': no matching filter type.", debug => 1);
             next;
         }
+        # Sometimes we may get such JSON: [{"lastResult":null,"name":null,"lastState":null,"type":null,"enabled":null}]
+        if (!defined($job->{name}) || $job->{name} eq 'null' ) {
```
garnier-quentin, line 165: null is not defined. you can remove the eq 'null' condition

garnier-quentin (dismissed):

apps/backup/veeam/local/mode/tapejobs.pm
Sims24: belt and braces! 😹 

Indeed ... thanks

garnier-quentin (approved):

PR #3332, opened 2021-12-21
fix(velocloud): compliant post request for qoe list

thibaults-centreon:
This fix the post request for qoe links where latest versions of Velocloud API won't accept quoted strings as values

### DIFF
--- a/cloud/vmware/velocloud/restapi/custom/api.pm
+++ b/cloud/vmware/velocloud/restapi/custom/api.pm
@@ -415,9 +415,9 @@ sub get_links_qoe {
         query_form_post => {
             enterpriseId => int($self->{entreprise_id}),
             edgeId => $options{edge_id},
-            debug => 'false',
-            individualScores => 'false',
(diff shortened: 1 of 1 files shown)


cloud/vmware/velocloud/restapi/custom/api.pm
```diff
@@ -415,9 +415,9 @@ sub get_links_qoe {
         query_form_post => {
             enterpriseId => int($self->{entreprise_id}),
             edgeId => $options{edge_id},
-            debug => 'false',
-            individualScores => 'false',
-            maxSamples => '15',
+            debug => \0,
```
garnier-quentin, line 418: You cannot set directly 0 ?

thibaults-centreon: this format is based on this reference:
https://metacpan.org/pod/JSON::XS#other-references

garnier-quentin (approved):

PR #3339, opened 2021-12-23
(plugin): os-windows-wmi - initial release (WIP)

Tpo76:
init plugin
wmi through wmic


Sims24:
Hello Fab, 

I think we should use Win::OLE instead of wmic binary.

PR #3443, opened 2022-01-28
enh(wsman): os-windows-wsman-plugin

Tpo76:
new modes os-windows-wsman-plugin


garnier-quentin:
https://github.com/centreon/centreon-plugins/pull/3451

PR #3446, opened 2022-01-31
(plugin) cloud::azure::network::trafficmanager - new plugin

matoy:


cloud/azure/network/trafficmanager/mode/queries.pm
```diff
@@ -0,0 +1,143 @@
...
+use warnings;
+
+sub get_metrics_mapping {
+    my ($self, %options) = @_;
+
+    my $metrics_mapping = {
+        'QpsByEndpoint' => {
+            'output' => 'Queries per second By Endpoint Count',
+            'label'  => 'queries-persecond-count',
+            'nlabel' => 'trafficmanager.queries.persecond.count',
```
Sims24, line 35: persecond is enough, we usually don't use both.

Sims24 (changes requested):
Thanks once again @matoy few things to change to merge this one.

matoy:
done!

Sims24:
Thank you!

Sims24:
@matoy one last thing, I checked aggregation Microsoft highlights for those metrics and I see that you use Average instead 
image

Any reason for that?

matoy:
no ; just a mistake.
actually average, min ou max would no change the result very much. but total should not be used (works but no interest in this value).

Sims24 (approved):

PR #3485, opened 2022-02-11
(plugin) apps::pvx::restapi - add default-value parameter

Sims24:
Sometimes users want to set a default value when no data is returned by PVX (Some kind of --zeroed parameter). 

Here we allow setting any value the user might want to set to identify that PVX didn't return any data back.

### DIFF
--- a/apps/pvx/restapi/custom/api.pm
+++ b/apps/pvx/restapi/custom/api.pm
@@ -44,18 +44,19 @@ sub new {
 
     if (!defined($options{noptions})) {
         $options{options}->add_options(arguments => {
-            'api-key:s'   => { name => 'api_key' },
-            'hostname:s'  => { name => 'hostname' },
-            'url-path:s'  => { name => 'url_path' },
-            'port:s'      => { name => 'port' },
-            'proto:s'     => { name => 'proto' },
-            'credentials' => { name => 'credentials' },
-            'basic'       => { name => 'basic' },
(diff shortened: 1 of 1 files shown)


apps/pvx/restapi/custom/api.pm
```diff
@@ -166,6 +167,13 @@ sub query_range {
 
     my $result = $self->get_endpoint(url_path => '/query?expr=' . $uri->encode($query));
 
+    if (defined( $self->{option_results}->{default_value} ) && !exists( $result->{data} )) {
+        my $forced_hash;
+        $options{filter} =~ /([^\s\\]+) = \"([^\s\\]+)\"/;
```
garnier-quentin, line 172: $options{filter} can be undefined no ?

apps/pvx/restapi/custom/api.pm
```diff
@@ -166,6 +167,13 @@ sub query_range {
...
+        push @{ $forced_hash->{key} }, { value => $2 };
+        push @{ $forced_hash->{values} }, { value => $self->{option_results}->{default_value} }; push @{$result->{data}}, $forced_hash;
```
garnier-quentin, line 174: i don't understand your data structure with forced_hash. You put it directly in result->data

Sims24: Yes indeed

PR #3532, opened 2022-03-08
enh(build): [WP] Debian package building with Jenkins

lgcosta:
Initial implementation process in production environment


garnier-quentin:
Sorry but we won't put that file here. Please use files from centreon-plugin-packs repository. We should use the same repository for centos and debian.

PR #3556, opened 2022-03-17
(core) snmp_standard::mode::storage - allocation units 0

garnier-quentin:


poomystery:
Hi,
 @garnier-quentin i tried your solution but it isnt working for me as i am facing the same issue 
even after making the changes, the output is still UNKNOWN: Illegal division by zero at /usr/lib/centreon/plugins//centreon_linux_snmp.pl line 12759

Sims24:
@poomystery please update the script from unstable repositories. Like this: 

`yum install centreon-plugin-Operatingsystems-Linux-Snmp --enablerepo='centreon-unstable*'`

PR #3587, opened 2022-04-11
(plugin) storage::netapp::ontap - add aggregates mode 

gaelatcheops:
Creation of Netapp plugins to observe aggregate status and latency.
Freely inspired by volumes plugins


garnier-quentin:
Could you provide the curl data results ?

PR #3606, opened 2022-04-19
fix(windows-wsman): list-storages backslash issue for GUI

Tpo76:
Substitute "\" for service discovery rules


garnier-quentin:
It works with the plugin. I think you have an isssue with autodisco or your rule.

PR #3856, opened 2022-08-22
(plugin) cloud::azure::recovery::plugin - adding mode replication-health

lchrdn:
Adding mode to Azure recovery plugin to check replication health and failover state.


lchrdn:
missing elements

PR #3888, opened 2022-09-08
(plugin) cloud::azure::management::monitor - fixing discovery tenant issues

lchrdn:
Fixing issue with subscription ID parsing 
Adding option to filter on resource types
Adding tags[ ] when it doesn't exist

### DIFF
--- a/cloud/azure/management/monitor/mode/discoverytenant.pm
+++ b/cloud/azure/management/monitor/mode/discoverytenant.pm
@@ -77,7 +77,10 @@ sub run {
                 }
                 $resource->{$entry} = \@array;
             }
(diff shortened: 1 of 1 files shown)


cloud/azure/management/monitor/mode/discoverytenant.pm
```diff
@@ -63,6 +64,8 @@ sub run {
         );
 
         foreach my $resource (@{$resources}) {
+            next if ($resource->{type} !~ /$self->{option_results}->{select_type}/i);
```
Sims24, line 67: check if select_type option is defined and ne ''

Sims24 (approved):

PR #3900, opened 2022-09-15
(plugin) os::linux::local::mode::pendingupdates - adding pending security updates metric 

lchrdn:
Adding new metric displaying how many security updates are available.

### DIFF
--- a/os/linux/local/mode/pendingupdates.pm
+++ b/os/linux/local/mode/pendingupdates.pm
@@ -119,8 +119,13 @@ sub parse_check_update {
     foreach my $line (@lines) {
         next if ($line !~ /^(\S+)\s+(\d+\S+)\s+(\S+)/
             && $line !~ /\s+(\S+)\s+\(\S+\s\=\>\s(\S+)\)/
-            && $line !~ /.*\|.*\|\s+(\S+)\s+\|.*\|\s+(\d+\S+)\s+\|.*/);
+            && $line !~ /.*\|.*\|\s+(\S+)\s+\|.*\|\s+(\d+\S+)\s+\|.*/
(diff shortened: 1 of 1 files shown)


os/linux/local/mode/pendingupdates.pm
```diff
@@ -68,10 +76,11 @@ sub custom_updates_output {
 
 sub new {
     my ($class, %options) = @_;
-    my $self = $class->SUPER::new(package => __PACKAGE__, %options);
+    my $self = $class->SUPER::new(package => __PACKAGE__, %options, force_new_perfdata => 1);
     bless $self, $class;
 
     $options{options}->add_options(arguments => {
+        'check-security:s'    => { name => 'check_security' },
```
Sims24, line 83: You can use a flag option, not a string (remove :s)

os/linux/local/mode/pendingupdates.pm
```diff
@@ -102,23 +112,20 @@ sub check_options {
     }    
 }
```
Sims24, line 114: It would be cool to directly return an UNKNOWN with a clear message explaining that --check-security is only available with rhel.  (if defined check_security && os mode not rhel ...)

os/linux/local/mode/pendingupdates.pm
```diff
@@ -142,6 +149,43 @@ sub manage_selection {
...
+sub manage_selection {
+    my ($self, %options) = @_;
+
+    my ($stdout) = $options{custom}->execute_command(
+        command => $self->{command},
+        command_options => $self->{command_options},
+        no_quit => 1
+    );
+
+    if ((defined($self->{option_results}->{check_security} && $self->{option_results}->{check_security} ne '') &&
```
Sims24, line 172: using a flag option, (see first comment) will save you the ne '' evaluation

Sims24 (approved):

PR #4024, opened 2022-10-26
(plugin) cloud::azure::management::cost - adding mode to query subscription costs

lchrdn:
Adding mode to check subscriptions costs, per subscription and resource groups. 
It is also possible to filter per tags.

### DIFF
--- a/cloud/azure/management/costs/mode/costsexplorer.pm
+++ b/cloud/azure/management/costs/mode/costsexplorer.pm
@@ -207,7 +207,7 @@ sub manage_selection {
             currency => $currency
         };
     }
-    if (defined($self->{option_results}->{resource_group}) || defined($self->{option_results}->{tags})){
+    if (defined($self->{option_results}->{resource_group}) || defined($self->{option_results}->{tags}) && keys %{$self->{tags}} > 0){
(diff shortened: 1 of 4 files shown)


cloud/azure/management/costs/mode/costsexplorer.pm
```diff
@@ -0,0 +1,290 @@
...
+
+sub check_options {
+    my ($self, %options) = @_;
+    $self->SUPER::check_options(%options);
+
+    if (defined($self->{option_results}->{tags}) && $self->{option_results}->{tags} ne '') {
+        foreach my $tag_pair (@{$self->{option_results}->{tags}}) {
+            my ($key, $value) = split / => /, $tag_pair;
+            $value = "" if !defined($value);
+            centreon::plugins::misc::trim($value) if defined($value);
```
Sims24, line 101: Not sure when $value can be undefined then. Line 101 condition sounds weird but I'm probably missing stg

lchrdn: Right. It cannot be undefined

Sims24 (approved):

PR #4036, opened 2022-11-04
(plugin) apps::vmware::connector - adding filter on alarm types for alarm-datacenter 

lchrdn:
Adding filter to exclude specified alarms types.

### DIFF
--- a/apps/vmware/connector/mode/alarmdatacenter.pm
+++ b/apps/vmware/connector/mode/alarmdatacenter.pm
@@ -182,6 +182,7 @@ sub new {
         'datacenter:s'  => { name => 'datacenter' },
         'filter'        => { name => 'filter' },
         'filter-time:s' => { name => 'filter_time' },
+        'filter-type'   => { name => 'filter_type' },
         'memory'        => { name => 'memory' }
     });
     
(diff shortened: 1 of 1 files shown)


apps/vmware/connector/mode/alarmdatacenter.pm
```diff
@@ -223,6 +224,8 @@ sub manage_selection {
         $self->{datacenter}->{$datacenter_name} = { display => $datacenter_name, alarm => {}, dc_metrics => { 1 => { name => $datacenter_name } } };
         
         foreach (keys %{$response->{data}->{$datacenter_id}->{alarms}}) {
+            next if (defined($self->{option_results}->{filter_type}) && $self->{option_results}->{filter_type} ne '' &&
+                $response->{data}->{$datacenter_id}->{alarms}->{$_}->{type} =~ /$self->{option_results}->{filter_type}/);
```
garnier-quentin, line 228: To get the same logic, it should be ```!=```

lchrdn: okay I thought we wanted to exclude, but yes we can keep only the alarms of specified type(s)

garnier-quentin: You can add an option to exclude also

Sims24 (approved):

PR #4152, opened 2023-01-18
pfsense : add states counter in packetstats module

kduret:
## Description

add states counter

migrated from https://github.com/centreon/centreon-plugins/pull/4082

## Type of change


garnier-quentin:
https://github.com/centreon/centreon-plugins/pull/4169

PR #4162, opened 2023-01-18
(plugin) apps::backup::rubrik::restapi - add service account auth

garnier-quentin:


itoussies (approved):

jonans91:
Hello. 
While the update is available and well applied on CentOS, the update is not available on debian (20230215-125031 appears to be the last one, but on CentOS 20230315-163435). 
How can we get the update on Debian ? 

Thanks

PR #4184, opened 2023-01-27
chore(ci): package plugins to el9

kduret:
## Description

package plugins to el9

**Fixes** MON-16456

## Type of change

## Target serie

- [ ] 21.10.x
- [ ] 22.04.x
- [ ] 22.10.x
- [x] 23.04.x (master)

<h2> How this pull request can be tested ? </h2>

Check github actions

### DIFF
--- a/.github/workflows/plugin-delivery.yml
+++ b/.github/workflows/plugin-delivery.yml
@@ -36,7 +36,8 @@ jobs:
     runs-on: [self-hosted, common]
     strategy:
       matrix:
-        distrib: [el7, el8]
+        distrib: [el9]
+        #distrib: [el7, el8, el9]
 
     steps:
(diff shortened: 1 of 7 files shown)


.github/workflows/plugin-delivery.yml
```diff
@@ -36,7 +36,7 @@ jobs:
     runs-on: [self-hosted, common]
```
sc979, line 36: Shouldn't we also change it to ubuntu

sc979 (approved):

.github/workflows/plugin-delivery.yml
kduret: Soon I guess
But currently delivery is possible only from internal network

PR #4334, opened 2023-04-08
(plugin) apps::protocols::x509 - add support for verify_hostname

julthomas:
This patch allows to get an UNKNOWN state when the following conditions are all true:

- plugin custommode tcp is used
- SSL_verify_mode => SSL_VERIFY_NONE is used (above all we check certicate expiration here)
- the CN/SANs in the certificate returned by the server does not match the servername given in the plugin options

It allows to check for certificate expiration and ensures the certificate presented by the server matches the name requested. Otherwise a valid certificate (dates) for an invalid servername would return an OK state, which ones can interpret as a wrong status considering the given servername option.

```
/opt/centreon-plugins/centreon_plugins.pl \
    --plugin 'apps::protocols::x509::plugin' \
    --mode 'certificate' \
    --custommode 'tcp' \
    --ssl-opt 'SSL_verify_mode => SSL_VERIFY_NONE' \
    --ssl-ignore-errors \
    --hostname '1.2.3.4' \
    --port '443' \
    --servername 'www.domain.tld' \
    --warning-status '%{expiration} < 45' \
    --critical-status '%{expiration} < 25' \
    --timeout '5'
```

### DIFF
--- a/src/apps/protocols/x509/custom/tcp.pm
+++ b/src/apps/protocols/x509/custom/tcp.pm
@@ -209,6 +209,21 @@ sub get_certificate_informations {
     $cert_infos->{subject} = $socket->peer_certificate('commonName');
     $cert_infos->{issuer} = $socket->peer_certificate('authority');
 
+    if (defined($self->{ssl_context}->{SSL_verify_mode}) &&
+        defined($self->{option_results}->{servername}) &&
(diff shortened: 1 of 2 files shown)


garnier-quentin:
https://github.com/centreon/centreon-plugins/pull/4364

PR #4384, opened 2023-04-24
(plugin) cloud::aws - fix paws custommode

garnier-quentin:


garnier-quentin:
https://github.com/centreon/centreon-plugins/pull/4388

PR #4385, opened 2023-04-24
apps::monitoring::kadiska::plugin - fix metric format

thibaults-centreon:
Kadiska API can return values with 8+ decimals, which is not pretty. 
This aims to format *time* metrics perfdatas as .2f instead of s


SchneiderNoel:
Hi,
Please note it's important to set .2f instead of s on many metrics in this plugin because there are many metrics with 8+ decimals, which is not pretty ! Not only for the time metrics perfdatas !

The .2f instead s in mandatory for the following metrics :
- watcher.dtt.spent.time.milliseconds
- watcher.errors.percentage;
- watcher.network.spent.time.milliseconds
- watcher.srt.spent.time.milliseconds
- watcher.waiting.time.milliseconds

Thx for your help on this.

garnier-quentin (approved):

PR #4417, opened 2023-05-16
(plugin) apps::backup::veeam::wsman - new 

Tpo76:
Work in progress
Veeam mode through wsman

EDIT by @omercier 
REFS: CTOR-304


garnier-quentin:
Veeam powershell are quite slow. I don't really want to push that. A new plugin will come and we use the API of the veeam manager.

PR #4429, opened 2023-05-25
fortimail plugin based on a FortiMail 400F Model

rmorandell-pgum:
Modes Available:
   cpu
   diskusage
   ha
   hardware
   interfaces
   list-interfaces
   list-queues
   load
   memory
   queues
   uptime
   
 The cpu, load, memory and diskusage are only % values. The MIB doesn't expose more.
 
image

i have done it using this FortiMail-400F from the snmpwalk. I inserted a general hardware sensor mode because the MIB doesn't give much more info

image


garnier-quentin:
Could you force the perfdata for modes interfaces and uptime ? (like we do with microsens for example)

rmorandell-pgum:
> Could you force the perfdata for modes interfaces and uptime ? (like we do with microsens for example)

ok, done.

how it works ? do I have to close this pullrequet and make a new one ?

garnier-quentin:
https://github.com/centreon/centreon-plugins/pull/4484

PR #4443, opened 2023-06-01
(plugin) apps::protocols::snmp - add snmp cache system

garnier-quentin:


src/apps/protocols/snmp/mode/cache.pm
```diff
@@ -0,0 +1,115 @@
...
+
+=head1 MODE
+
+Cache SNMP datas in a JSON cache file.
+
+=over 8
+
+=item B<--file>
+
+JSON cache file.
```
itoussies, line 103:
```suggestion
JSON cache file path.
```

src/centreon/plugins/snmp.pm
```diff
@@ -1023,6 +1144,10 @@ Auto reduce SNMP request size in case of SNMP errors (By default, the divisor is
 
 Use snmp getnext function (even in snmp v2c and v3).
 
+=item B<--snmp-cache-file>
+
+Use SNMP cache file.
```
itoussies, line 1149: Maybe here we can say how to do it ? Or just something like "use this mode to do that" ?

PR #4449, opened 2023-06-02
(plugin) cloud::aws::cloudtrail - new

sdepassio:


packaging/centreon-plugin-Cloud-Aws-Cloudtrail-Api/rpm.json
```diff
@@ -0,0 +1,6 @@
+{
+    "dependencies": [
+        "perl(DateTime)"
+    ],
+    "custom_pkg_data": "Obsoletes:    centreon-plugin-Cloud-Aws-Api"
```
garnier-quentin, line 5: Why that obsolete configuration ?

src/cloud/aws/custom/paws.pm
```diff
@@ -839,6 +840,68 @@ sub directconnect_describe_virtual_interfaces {
     return $results;
 }
 
+sub cloudtrail_events {
+    my ($self, %options) = @_;
+
+    my $events_results = [];
+    eval {
+        my $ct;
+        if (length $self->{option_results}->{endpoint}) {
```
garnier-quentin, line 849: it can be undefined. that syntax can do some undefined value with older perl version

src/cloud/aws/custom/paws.pm
```diff
@@ -839,6 +840,68 @@ sub directconnect_describe_virtual_interfaces {
...
+        }
+        my %ct_options = ();
+        if (defined($options{delta})) {
+            $ct_options{EndTime} = time();
+            $ct_options{StartTime} = $ct_options{EndTime} - ($options{delta} * 60);
+        }
+
+        while (my $list_events = $ct->LookupEvents(%ct_options)) {
+            foreach (@{$list_events->{Events}}) {
+                my $event = JSON::XS->new->utf8->decode($_->{CloudTrailEvent});
```
garnier-quentin, line 862: i think you don't need to json decode.

Why you don't use the same method ? https://docs.aws.amazon.com/cli/latest/reference/cloudtrail/get-trail-status.html

sdepassio: it doesn't work without the json decode because (it is a json string as a value in the json returned)

garnier-quentin: indeed

garnier-quentin (approved):

PR #4495, opened 2023-06-21
(plugin) os::linux::local - fix improper AIX field descriptor

garnier-quentin:


philipraets:
Just tried the patch, and it is working

thx,

Philip

itoussies (approved):

PR #4517, opened 2023-06-28
enh(plugin): update email notification mode

cgagnaire:
Update the mode so : 
 - the message can be shown in Outlook (if condition not closed, and useless),
 - an empty attachment is removed for host notification,
 - the graph (or No Graph string) is centered,
 - the background around the body is grey (sexier).


garnier-quentin:
https://github.com/centreon/centreon-plugins/pull/4520

PR #4518, opened 2023-06-29
Emerson/Vertiv energy station plugin

rmorandell-pgum:
Modes Available:
   battery
   input
   rectifier
   system
   temperature
   uptime

battery
> voltage
> current
> capacity
> used capacity
> battery mode

input
> line-A
> line-B
> line-C

rectifier
> status
> used-capacity
> installed
> communicating

system
> status (system_status, communication_status)
> current
> installed
> used-capacity

system
> alarm_status
> temperature

### DIFF
--- a/src/hardware/ups/ees/vertiv/snmp/mode/battery.pm
+++ b/src/hardware/ups/ees/vertiv/snmp/mode/battery.pm
@@ -42,23 +42,12 @@ my $map_battery_mode = {
     13 => 'MasterBateryTesting',
 };
 
-my $thresholds = {
-    'battery-mode' => [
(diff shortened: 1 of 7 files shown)


garnier-quentin:
I have checked and it's an UPS. maybe we can put it in devices

rmorandell-pgum:
Hi .. i don't understand. yes it's an UPS

i made hardware::ups::ees::vertiv::snmp::plugin

it's that not the correct path for UPS ?

garnier-quentin:
you're right. But on the site, the constructor doesn't use the term of UPS. maybe for us it's more than an UPS.

PR #4536, opened 2023-07-06
enh(plugin): better handle when graph cannot be retrieved

cgagnaire:
Better handle when graph cannot be retrieved, and show why.


garnier-quentin:
thanks @cgagnaire https://github.com/centreon/centreon-plugins/pull/4544

PR #4545, opened 2023-07-07
(plugin) apps::monitoring::iplabel::newtest::restapi - mode scenarios…

garnier-quentin:
… add --timeframe option

### DIFF
--- a/src/apps/monitoring/iplabel/newtest/restapi/mode/scenarios.pm
+++ b/src/apps/monitoring/iplabel/newtest/restapi/mode/scenarios.pm
@@ -307,7 +307,8 @@ Filter scenarios (can be a regexp).
 
 =item B<--timeframe>
 
-Define timeframe in seconds (by default it's the time between two executions).
+Define timeframe (duration to go back in time) in seconds (default value: 300 ).
+The first range is calculated with this value then it's calculated from the last execution (stored in the cache file).
 
 =item B<--warning-*> B<--critical-*>
 
--- a/src/apps/monitoring/iplabel/newtest/restapi/mode/scenarios.pm
+++ b/src/apps/monitoring/iplabel/newtest/restapi/mode/scenarios.pm
@@ -140,8 +140,6 @@ sub manage_selection {
     } else {
         $last_timestamp = time() - (60 * $timespan);
     }
-    
-    
 
     my $instances = $options{custom}->request_api(endpoint => '/api/instances');
     $self->{robots} = {};
--- a/src/apps/monitoring/iplabel/newtest/restapi/mode/scenarios.pm
+++ b/src/apps/monitoring/iplabel/newtest/restapi/mode/scenarios.pm
@@ -115,7 +115,8 @@ sub new {
 
     $options{options}->add_options(arguments => { 
         'filter-robot-name:s'    => { name => 'filter_robot_name' },
-        'filter-scenario-name:s' => { name => 'filter_scenario_name' }
+        'filter-scenario-name:s' => { name => 'filter_scenario_name' },
+        'timeframe:s'            => { name => 'timeframe' }
     });
 
     return $self;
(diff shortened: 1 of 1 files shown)


src/apps/monitoring/iplabel/newtest/restapi/mode/scenarios.pm
```diff
@@ -130,10 +131,14 @@ sub manage_selection {
...
     my $timespan = 5;
+    if (defined($self->{option_results}->{timeframe}) && $self->{option_results}->{timeframe} =~ /^(\d+)$/) {
+        $timespan = $1 / 60;
+    }
+
     if (defined($last_timestamp)) {
         $timespan = POSIX::ceil((time() - $last_timestamp) / 60);
     } else {
-        $last_timestamp = time() - (60 * 5);
+        $last_timestamp = time() - (60 * $timespan);
```
lucie-dubrunfaut, line 141: At this time the code checks at each execution if the timeframe option is defined by user even if 
$last_timestamp can be calculated from cache file. It may be more relevent to avoid checking the 
values if not needed.
After reading ticket, it remain unclear for me if the user wants to set timeframe option for first 
execution only or for all executions. I suggest both cases below. And also I tried to uniformize 
times in second for $timespan definition, removing the POSIX::ceil call too.

First case : First range is calculated with default value or with timeframe option set by user
then it's calculated from the timestamp present in the cache file.

```
    my $last_timestamp = $self->read_statefile_key(key => 'last_timestamp');

    my $timespan = 300; # default value: 5 min (only use if not set by user at the first time)

    if (defined($last_timestamp)) {
        $timespan = time() - $last_timestamp;
    }else{
    	if (defined($self->{option_results}->{timeframe}) && $self->{option_results}->{timeframe} =~ /^(\d+)$/) {
             $timespan = $1;
        }
        $last_timestamp = time() - $timespan;
    }
```

Second case : User wants to use at every execution the same range set in timeframe.
If not, first range is calculated with default value and then it's calculated from 
the timestamp present in the cache file.

```
    my $last_timestamp = $self->read_statefile_key(key => 'last_timestamp');

    my $timespan = 300; # default value: 5 min (only use if not set by user at the first time)

    if (defined($self->{option_results}->{timeframe}) && $self->{option_results}->{timeframe} =~ /^(\d+)$/) {
        $timespan = $1;
        $last_timestamp = time() - $timespan;
    }else{
        if (defined($last_timestamp)) {
             $timespan = time() - $last_timestamp;
        } else {
             $last_timestamp = time() - $timespan;
        }
    }
```

src/apps/monitoring/iplabel/newtest/restapi/mode/scenarios.pm
```diff
@@ -299,6 +304,10 @@ Filter robots (can be a regexp).
 
 Filter scenarios (can be a regexp).
 
+=item B<--timeframe>
+
+Define timeframe in seconds (by default it's the time between two executions).
```
lucie-dubrunfaut, line 309: We should add timeframe default value here

`Define timeframe (time between two executions) in seconds (default value: 300 ).`

lucie-dubrunfaut: Is this default value explanation wrong or anything else ?

sdepassio (approved):

src/apps/monitoring/iplabel/newtest/restapi/mode/scenarios.pm
```diff
@@ -299,6 +305,10 @@ Filter robots (can be a regexp).
...
```
lucie-dubrunfaut, line 310: Define timeframe (time between two executions) in seconds (default value: 300 ).
The first range is calculated with this value then it's calculated from the timestamp present in the cache file.

lucie-dubrunfaut (changes requested):

src/apps/monitoring/iplabel/newtest/restapi/mode/scenarios.pm
omercier, line 310:
```suggestion
Define timeframe (duration to go back in time) in seconds (default value: 300 ).
The first range is calculated with this value then it's calculated from the last execution (stored in the cache file).
```

omercier (changes requested):

lucie-dubrunfaut (approved):

omercier (approved):

PR #4566, opened 2023-07-19
(plugin) apps::backup::rubrik::restapi - mode jobs add cache and limit

garnier-quentin:


itoussies (approved):

jonans91:
Hello (Bonjour Quentin!)

Since the update of the script and the mode job added, there is a problem with the service account. 
The authentication works well during a few minutes / hours, and then poller can't not connect properly to the host Rubrik. Error 422 (de mémoire)

I can reproduce the problem with the latest centreon_rubrik script. When i erase it with the old one (before july), no problem, all is working fine. 

Available for test if needed.

PR #4575, opened 2023-07-24
new(plugin): add docker local/ssh plugin

cgagnaire:
Add Docker plugin using docker cli (local or SSH).
This way you don't use the restapi/containerd setup which exposes all endpoints to a root-like privilege.
But, because the cli scales values for statistics this is far less precise.

### DIFF
--- a/src/cloud/docker/local/custom/cli.pm
+++ b/src/cloud/docker/local/custom/cli.pm
@@ -1,0 +1,194 @@
+#
+# Copyright 2023 Centreon (http://www.centreon.com/)
+#
+# Centreon is a full-fledged industry-strength solution that meets
+# the needs in IT infrastructure and application monitoring for
+# service performance.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+package cloud::docker::local::custom::cli;
+
+use strict;
+use warnings;
+use centreon::plugins::ssh;
+use centreon::plugins::misc;
+
+sub new {
+    my ($class, %options) = @_;
+    my $self  = {};
+    bless $self, $class;
+
+    if (!defined($options{output})) {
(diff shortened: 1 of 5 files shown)


garnier-quentin:
thanks @cgagnaire could you provide command and output examples ?

cgagnaire:
```
# perl centreon_plugins.pl --plugin=cloud::docker::local::plugin --mode=container-usage --hostname=myhostname --ssh-username=sup --ssh-priv-key=/var/lib/centreon-engine/.ssh/sup --ssh-backend=libssh --sudo --filter-name=mariadb
OK: Container 'mariadb' cpu usage: 0.01 %, memory total: 7.79 GB used: 13.94 MB (0.17%) free: 7.78 GB (99.83%), disk read throughput: 0.00 B/s, disk write throughput: 0.00 B/s, traffic in: 0.00 b/s, traffic out: 0.00 b/s | 'mariadb#container.cpu.utilization.percentage'=0.01%;;;0;100 'mariadb#container.memory.usage.bytes'=14617149B;;;0;8364448808 'mariadb#container.disk.throughput.read.bytespersecond'=0B/s;;;0; 'mariadb#container.disk.throughput.write.bytespersecond'=0B/s;;;0; 'mariadb#container.traffic.in.bitspersecond'=0b/s;;;0; 'mariadb#container.traffic.out.bitspersecond'=0b/s;;;0;
```

```
# perl centreon_plugins.pl --plugin=cloud::docker::local::plugin --mode=container-status --hostname=myhostname --ssh-username=sup --ssh-priv-key=/var/lib/centreon-engine/.ssh/sup --ssh-backend=libssh --sudo --filter-name=mariadb
OK: Container 'mariadb' status is 'Up 28 days' 
```

```
# perl centreon_plugins.pl --plugin=cloud::docker::local::plugin --mode=list-containers --hostname=myhostname --ssh-username=sup --ssh-priv-key=/var/lib/centreon-engine/.ssh/sup --ssh-backend=libssh --sudo 
List containers: 
[id = 0b4af843a1ef] [name = mariadb] [status = Up 28 days]
[id = 54f023b32e40] [name = apache2] [status = Up 2 days]
[id = 432bde4f3452] [name = loki] [status = Up 1 days]
```

garnier-quentin:
@cgagnaire have you tried the option [--format](https://docs.docker.com/engine/reference/commandline/stats/#format)  to have a json output ?

cgagnaire:
Yes but the values are the same as the table format and its a "per line" json so there is no real gain.
And also sudoers conf does not seem to like curly braces so it would be counterproductive if you cannot define the all command in the conf.

garnier-quentin:
The new PR: https://github.com/centreon/centreon-plugins/pull/4580

Could you test it ? I have done a minor change

PR #4584, opened 2023-07-28
fix(pg/pkg): fix postgresql packaging dependencies

cgagnaire:
Add missing perl dependencies for Postgresql plugin packaging


garnier-quentin:
thanks @cgagnaire https://github.com/centreon/centreon-plugins/pull/4585

PR #4590, opened 2023-08-04
enh(http/collection): add second2human function

cgagnaire:
Add a second2human function to translate a duration in seconds to a human readable string.


garnier-quentin:
thanks @cgagnaire https://github.com/centreon/centreon-plugins/pull/4610

PR #4594, opened 2023-08-07
enh(centreon/sql): add executiontime mode

cgagnaire:
Simple mode to count and list services for which execution time exceeds a defined amount of time.


garnier-quentin:
thanks @cgagnaire https://github.com/centreon/centreon-plugins/pull/4614

PR #4651, opened 2023-09-09
performance data: add unit ms to jitter metric

joschi99:
This will add unit ms for jitter performance data


garnier-quentin:
https://github.com/centreon/centreon-plugins/pull/4664

PR #4683, opened 2023-10-04
fix(docker/local): fix cache name to prevent conflict

cgagnaire:
Fix the cache name to prevent conflict between containers with same being check from different hosts.

EDIT by @omercier 
REFS: MON-22340


lucie-dubrunfaut:
thanks @cgagnaire : https://github.com/centreon/centreon-plugins/pull/4690

PR #4684, opened 2023-10-04
enh(http/collection): add returned json structure in debug output

cgagnaire:
Add a pretty JSON structure from the returned body as part of the debug message to help users parse the result with jpath method (very useful when body is actually in XML).

EDIT by @omercier 
REFS: MON-22338


lucie-dubrunfaut:
Add in MON-22338 : https://github.com/centreon/centreon-plugins/pull/4708

Issue #4645, opened 2023-09-07
(vmware/health) odd dealing of ok problems


cgagnaire:
Hey @garnier-quentin,
I think there may be a bug on how the problems are counted in the health mode.
If I check a host, I'll have:
OK: 0 total health issue(s) found - Host 'toto' : status connected - 173 health checks are green, **173 total health issue(s) found**, 0 yellow health issue(s) found, 0 red health issue(s) found - temperature sensors are ok - fan sensors are ok - voltage sensors are ok
I think we should increment total_problems counter only if the sensor has a problem right ? 
https://github.com/centreon/centreon-plugins/blob/cf2fa4366bdfaf7bdae9933837f59bd936ebda27/src/apps/vmware/connector/mode/healthhost.pm#L261
The total counter is incremented 2 lines before as it sould.

EDIT by @omercier 
REFS: MON-21885

garnier-quentin:
there is an issue you're right

garnier-quentin:
It should be something like that:
```
                 if ($status ne 'ok') {
                        $self->{host}->{$host_name}->{global_problems}->{total_problems}++ 
                        $self->{host}->{$host_name}->{global_summary}->{$i} = {
                            type => defined($entry->{type}) ? $entry->{type} : '',
                            name => $entry->{name},
                            summary => $entry->{summary}
                        };
                    }
```

cgagnaire:
Yes seems good to me.

lucie-dubrunfaut:
Fix [here](https://github.com/centreon/centreon-plugins/pull/4716)

PR #4705, opened 2023-10-18
fix(vmware/health): fix health counting ok problems

cgagnaire:
Count problems when status is not ok.

Fixes https://github.com/centreon/centreon-plugins/issues/4645


lucie-dubrunfaut:
PR add [here](https://github.com/centreon/centreon-plugins/pull/4716)

PR #4744, opened 2023-11-17
fix: os::windows::snmp::mode::service not working with special italian or german language chars because filter not decoded

rmorandell-pgum:
## Description

When using some German or Italian systems the service names can contain some special language chars. The plugin is working fine because the value receiving from SNMP is UTF-8 decoded (when you don't use --source-encoding UTF-8 is the standard).

The problem is when use a filter to filter only one specific service the filter will not be decoded. So the plugin will filter with a regex that not matches because the filter is not decoded.

The check fails

image

image

**Fixes** # 

decode the filter, too and the check will work fine

image

and the check works

image

## Type of change

## Target serie

- [ ] 22.04.x
- [x] 22.10.x
- [x] 23.04.x
- [x] 23.10.x
- [x] 24.04.x (master)

<h2> How this pull request can be tested ? </h2>

You can change a Windows Service display name and check it via snmp.
sc config "servicename" displayname= "new displayname"

### DIFF
--- a/src/os/windows/snmp/mode/service.pm
+++ b/src/os/windows/snmp/mode/service.pm
@@ -212,6 +212,7 @@ sub manage_selection {
         my $instance = $1 . '.' . $2;
 
         my $name = $self->{output}->decode(join('', map(chr($_), split(/\./, $2))));
+        $self->{option_results}->{filter_name} = $self->{output}->decode($self->{option_results}->{filter_name});
         
         next if (defined($self->{option_results}->{filter_name}) && $self->{option_results}->{filter_name} ne '' &&
             $name !~ /$self->{option_results}->{filter_name}/);


rmorandell-pgum:
Hi @omercier,
for info: the  --list-services mode is doing it correctly

sudo -u centreon-engine /usr/lib/i-vertix/plugins/os-windows-snmp.pl --hostname=10.10.101.6 --snmp-community=**** --snmp-version=1 --snmp-timeout= --plugin=os::windows::snmp::plugin --mode=list-services --filter-name=test
List services:
[name = testservice üäöáà (123)] [operating = active] [installed = installed]

Attached the snmpwalk. It must been the service name where the name is the Hex String and not a normal string

[services.zip](https://github.com/centreon/centreon-plugins/files/13948899/services.zip).

i hope it helps. 

regards

omercier:
Hi @rmorandell-pgum,

Thanks for the contribution and for the additional details :100:

Your contribution will be discussed during the next weekly sessions of refinement with the team.

Thanks :+1:

Regards,

omercier:
Thanks again @rmorandell-pgum ! Your commit is now in PR #4955 and it will be part of the next release, planned on 11th april.
Best regards

PR #4850, opened 2024-01-19
enh(http/collection): move json parsing outside http call

cgagnaire:
This PR moves JSON (and XML) parsing outside the call_http function so we can actually use the scenario_stopped directive if we don't do recursive requests (only one request, which must be most use cases).

It was not possible before because the parsing bit was stopping the scenario for JSON decoding problem before HTTP metrics were used (through check_filter2 function).

Now, for example we can look at HTTP returned code before parsing a content, so we can exit the scenario in case of non-2xx.

This example adds a "test" selection that shows the returned code and message that can handle, in addition to the "scenario_stopped" directive, the cases where the API does not respond as expected. That way, we don't have "Cannot decode ..." unknown output when the API actually doesn't work.

```
{
    "http": {
        "requests": [
            {
                "name": "ping",
                ...
                "endpoint": "/toto",
                "method": "GET",
                ...
                "insecure": 0,
                "timeout": 30,
                ...
                "backend": "curl",
                "scenario_stopped": "%(builtin.httpCode.ping) != 200",
                "rtype": "xml",
                "parse": [
                    {
                        "name": "result",
                        "path": "$..PingResponse.PingResult",
                        "entries": [
                            {
                                "id": "ElapsedMilliseconds"
                            }
                        ]
                    }
                ]
            }
        ]
    },
    "selection": [
        {
            "name": "test",
            "critical": "%(builtin.httpCode.ping) != 200",
            "exit": "%(builtin.httpCode.ping) != 200",
            "formatting": {
                "printf_msg":"API returned HTTP code '%s' with message '%s'",
                "printf_var":[
                    "%(builtin.httpCode.ping)",
                    "%(builtin.httpMessage.ping)"
                ],
                "display_ok": true
            }
        },
        {
            "name": "status",
            "formatting": {
                "printf_msg": "Worker ping returned in %s ms",
                "printf_var": [
                    "%(http.tables.pingResult.[0].ElapsedMilliseconds)"
                ],
                "display_ok": true
            },
            "perfdatas": [
                {
                    "nlabel": "worker.ping.milliseconds",
                    "value": "%(http.tables.pingResult.[0].ElapsedMilliseconds)",
                    "min": 0,
                    "unit": "ms"
                }
            ]
        }
    ],
    "formatting": {
        "custom_message_global": "Saas platform returned calls successfully",
        "separator": "-"
    }
}
```

### DIFF
--- a/src/apps/protocols/http/mode/collection.pm
+++ b/src/apps/protocols/http/mode/collection.pm
@@ -87,7 +87,7 @@ sub custom_select_output {
 
     if (defined($format)) {
         return sprintf(
-            $format->{printf_msg}, @{$format->{printf_var}} 
+            $format->{printf_msg}, @{$format->{printf_var}}
         );
     }
 
@@ -123,7 +123,7 @@ sub new {
     my ($class, %options) = @_;
     my $self = $class->SUPER::new(package => __PACKAGE__, %options, force_new_perfdata => 1);
     bless $self, $class;
-    
+
     $options{options}->add_options(arguments => {
         'config:s'            => { name => 'config' },
         'filter-selection:s%' => { name => 'filter_selection' },
@@ -192,7 +192,7 @@ sub get_map_value {
     my ($self, %options) = @_;
 
     return undef if (
-        !defined($self->{config}->{mapping}) || 
+        !defined($self->{config}->{mapping}) ||
         !defined($self->{config}->{mapping}->{ $options{map} })
     );
     return '' if (!defined($self->{config}->{mapping}->{ $options{map} }->{ $options{value} }));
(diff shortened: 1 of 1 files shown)


omercier:
Hi @garnier-quentin and @cgagnaire,
First thanks to both of you for your contributions.
Do you think there is a way to combine both your fixes/enhancements?
If not, we'll first merge Quentin's PR to fix this infinite loop issue, that has proven useful to at least one of our customers.
Regards

garnier-quentin:
I'll check @cgagnaire last comment: https://github.com/centreon/centreon-plugins/pull/4915

garnier-quentin:
I have updated the patch @omercier

omercier:
Closing the PR (see #4915 and #5044).
Thanks again @cgagnaire !

PR #4861, opened 2024-01-23
enh(passwordmgr): handle different kind of options (string, arrays and hashes)

cgagnaire:
Adds support for the different kinds of options that a user can provide depending on the plugin/mode.
For examples Jolokia plugins are expecting arrays, Collections plugins are expecting key/value hash for constant options.


cgagnaire:
Doesn't handle centreonvault on purpose as it needs code changes on the Centreon side.

PR #4868, opened 2024-01-30
Fix API endpoint for K8s CronJob

poponealex:
**CronJob has moved from batch v1beta to v1 since kube 1.21**

The API endpoint has been updated in this PR to fix the 404 not found error when listing CronJobs using the Kubernetes API plugin in Centreon.


lucie-dubrunfaut:
Hello :)

Thanks for the contribution, this PR is duplicate from #5026 so let me close it and we'll handle this issue in the next future.

PR #4918, opened 2024-02-19
Ctor 378 fix snmp collections random issue

omercier:
## Description

I found a random issue that seems to happen when the source of the instances IDs is the same as one of the metrics'.
What I'm proposing here may only be a workaround, but it does not seem harmful to me.
My test performs 3 times the same command. If you run it without the fix, you will almost certainly reproduce this behaviour:

```
$ perl src/centreon_plugins.pl --plugin apps::protocols::snmp::plugin --mode collection --hostname 127.0.0.1 --snmp-port 2024 --snmp-community hardware-ups/hardware-ups-sputnik --config src/contrib/collection/snmp/sputnik-environment.json
Use of uninitialized value $1 in hash element at /home/olivier/projects/centreon-plugins/src/apps/protocols/snmp/mode/collection.pm line 248.
Use of uninitialized value $instance in hash element at /home/olivier/projects/centreon-plugins/src/apps/protocols/snmp/mode/collection.pm line 251.
OK: All selections are ok | '1#environment.temperature.celsius'=20.06C;;;; '1#environment.humidity.percent'=33%;;;0;100
``` 

Instead of this:

```
OK: Sensor '1' temperature is '20.06'°C and humidity is '33'% | '1#environment.temperature.celsius'=20.06C;;;; '1#environment.humidity.percent'=33%;;;0;100
```

## Type of change

### DIFF
--- a/tests/functional/snmp/snmp-collection-sputnik-snmp.robot
+++ b/tests/functional/snmp/snmp-collection-sputnik-snmp.robot
@@ -1,0 +1,46 @@
+*** Settings ***
+Documentation       Hardware UPS Sputnik SNMP plugin
+
+Library             OperatingSystem
+Library             String
+Library             Examples
+
+Test Timeout        120s
+
+
+*** Variables ***
+${CENTREON_PLUGINS}         ${CURDIR}${/}..${/}..${/}..${/}src${/}centreon_plugins.pl
+
+${CMD}                      perl ${CENTREON_PLUGINS} --plugin=apps::protocols::snmp::plugin
+
+*** Test Cases ***
+SNMP Collection - Sputnik Environment ${tc}/3
+    [Tags]    SNMP Collection
+    ${command}    Catenate
+    ...    ${CMD}
+    ...    --mode=collection
+    ...    --hostname=127.0.0.1
+    ...    --snmp-version=2c
+    ...    --snmp-port=2024
+    ...    --snmp-community=hardware-ups/hardware-ups-sputnik
+    ...    --config=../../../src/contrib/collection/snmp/sputnik-environment.json
+
+    ${output}    Run    ${command}
+    ${output}    Strip String    ${output}
+    Should Be Equal As Strings
+    ...    ${output}
+    ...    ${expected_result}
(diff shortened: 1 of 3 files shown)


kduret (dismissed):

.github/workflows/spellchecker.yml
```diff
@@ -48,6 +48,9 @@ jobs:
         run: |
           set +e
           for f in ${{ steps.filter.outputs.plugins_files }}; do
+            if [[ "$f" =~ \.json$ ]] ; then
+              continue
+            fi
```
sdepassio, line 53: I think it would be better to update the "filter" step to have only ".pm" files

omercier: Ok. I've pushed an attempt.
And added quotes, according to [this recommendation](https://github.com/dorny/paths-filter?tab=readme-ov-file#notes):
> It's recommended to quote your path expressions with ' or ". Otherwise, you will get an error if it starts with *.

sdepassio (approved):

kduret (approved):

tests/functional/snmp/snmp-collection-sputnik-snmp.robot
```diff
@@ -0,0 +1,46 @@
...
+    ...    ${expected_result}
+    ...    Wrong output result for compliance of ${expected_result}{\n}Command output:{\n}${output}{\n}{\n}{\n}
+
+    Examples:        tc    expected_result    --
+            ...      1     OK: Sensor '1' temperature is '20.06'°C and humidity is '33'% | '1#environment.temperature.celsius'=20.06C;;;; '1#environment.humidity.percent'=33%;;;0;100
+            ...      2     OK: Sensor '1' temperature is '20.06'°C and humidity is '33'% | '1#environment.temperature.celsius'=20.06C;;;; '1#environment.humidity.percent'=33%;;;0;100
+            ...      3     OK: Sensor '1' temperature is '20.06'°C and humidity is '33'% | '1#environment.temperature.celsius'=20.06C;;;; '1#environment.humidity.percent'=33%;;;0;100
+
+*** Keywords ***
+Append Option
```
Evan-Adam, line 41: I don't see any use of this keyword

PR #4953, opened 2024-03-15
Changed the 'experimental' message of Veeam WSMAN modules as we are going to support it

omercier:
## Description

Help messages modified.


.github/scripts/pod_spell_check.t
```diff
@@ -8,7 +8,7 @@ if (!@ARGV) {
     die "Missing perl file to check.";
 }
```
sc979, line 10: As the "pod_spell_check." script relates to perl, could you adapt the "CODE_OWNER" file and add it to the owners-perl's scope ?

sc979 (approved):

PR #5026, opened 2024-05-14
fix(plugin) kubenetes api

Tpo76:
Fix unrelevant path on latest k8s version for cronjob api and ingresses api
```
/usr/lib/centreon/plugins//centreon_kubernetes_api.pl --plugin=cloud::kubernetes::plugin --mode=list-ingresses --custommode='api' --hostname=172.16.5.223 --port='6443' --proto='https' --token='***' --config-file='' --proxyurl='' --namespace='default' --timeout='10'
List ingresses: 
[uid: 80b37c13-dea7-42cb-bb5d-e49cf0765d5c] [name: apache-ingress] [namespace: default]

/usr/lib/centreon/plugins//centreon_kubernetes_api.pl --plugin=cloud::kubernetes::plugin --mode=cronjob-status --custommode='api' --hostname=172.16.5.223 --port='6443' --proto='https' --token='***' --config-file='' --proxyurl='' --namespace='default' --timeout='10' --ssl-opt="SSL_verify_mode => SSL_VERIFY_NONE" --filter-name='.*' --filter-namespace='.*' --warning-status='' --critical-status='' --verbose 
OK: All CronJobs status are ok
```

CTOR-713


Evan-Adam:
do you know what kube version you client have and since which this path became unrevelent ?
we can't change the api we use if other client have an older version still active, that would force them to update.

Tpo76:
Client version is

```
Client Version: v1.30.0
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.30.0
```

No idea when it got officially release

Evan-Adam (dismissed):
https://kubernetes.io/blog/2021/04/09/kubernetes-release-1.21-cronjob-ga/
cronjob passed ga in 2021 (version 1.21) which is unsupported since 2022)
https://kubernetes.io/fr/releases/

garnier-quentin:
You need an option to override the path and be backward compatible

lucie-dubrunfaut:
Fixed with https://github.com/centreon/centreon-plugins/pull/5109 taking account of Quentin's comment.

PR #5168, opened 2024-09-05
HPE Primera API OPTIONS not shown in --help

rmorandell-pgum:
# Community contributors

## Description

The custom API options (like --hostname or --api-username) are not shown during --help on a mode of the plugin.

This happens because the section name in add_help function must be case sensitive.


omercier:
Thanks @rmorandell-pgum!

PR #5223, opened 2024-10-08
enh(docker::local): Make compatible with Podman containers

jensrudolf:
# Community contributors

## Description

Modify the cloud::docker::local::plugin to be compatible with Podman container engine.

Podman uses two whitespaces at minimum to seperate its cli output colums whereas Docker uses three. This reduces the required whichtespaces in the regex filter of the command output accordingly.

## Type of change

- [x] Functionality enhancement or optimization (non-breaking change)

Tested locally on RHEL8 with Icinga, Podman and Docker containers.

## Checklist

- [ ] I have commented my code, especially **hard-to-understand areas** of the PR.
- [ ] I have provide data or shown output displaying the result of this code in the plugin area concerned.

### DIFF
--- a/src/cloud/docker/local/mode/containerstatus.pm
+++ b/src/cloud/docker/local/mode/containerstatus.pm
@@ -79,7 +79,7 @@ sub manage_selection {
 
     shift(@lines);
     foreach my $line (@lines) {
-        next if ($line !~ /^(\S+)\s{3,}(\S+)\s{3,}(.*?)\s{3,}(.*?)\s{3,}(.*?)\s{3,}(.*?)\s{3,}(\S+)$/);
+        next if ($line !~ /^(\S+)\s{2,}(\S+)\s{2,}(.*?)\s{2,}(.*?)\s{2,}(.*?)\s{2,}(.*?)\s{2,}(\S+)$/);
 
         my ($id, $image, $command, $created, $status, $ports, $name) = ($1, $2, $3, $4, $5, $6, $7);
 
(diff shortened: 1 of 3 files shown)


lucie-dubrunfaut:
Hello :)

Thank you for your contribution. We plan to work very soon on a dedicated Podman plugin (you can vote on the [idea](https://thewatch.centreon.com/ideas/new-connector-plugin-pack-for-podman-2468) related in TheWatch). We would be delighted to discuss your specific requirements for a Podman connector with you in more detail.

jensrudolf:
Hi Lucie,

thanks for your reply, A dedicated Podman plugin sounds fantastic. Currently, we only monitor whether containers are running or not. This is already possible with this docker plugin if you provide a softlink docker -> podman and apply the patch in this MR to adapt the regex for the number of whitespaces podman uses to separate its output.

Another approach might be to replace the docker plugin with a more generic container plugin that supports both docker and podman via CLI or the REST API preferably with the local socket so the Docker/Podman API does not have to be exposed outside the monitored host.

What do you think?

BR
Jens

lucie-dubrunfaut:
Hello @jensrudolf :)

We'll try to take your suggestion into account, but I think that the Docker and Podman packs will be separated for maintenance reasons in the future if there's too much divergence between them. However, we have planned a common code base (as in the case of the `snmp_standard` plugin, which is called in a large number of plugins). 
About protocols, for the time being our development team will address the need for a Podman RestAPI plugin. It will be possible to adapt Cli afterwards (our plugins are designed to manage several protocols, generally via custom modes) but this is not planned for the time being (however, you can help us prioritise this need by creating a [TheWatch idea](https://thewatch.centreon.com/ideas) on the subject).

Best regards.

lucie-dubrunfaut:
Hello :)

We just released the new podman plugin with this PR https://github.com/centreon/centreon-plugins/pull/5427
Feel free to test it and give us feedbacks about it :)

PR #5256, opened 2024-10-23
Fix: HTTP status handling & retry on invalid token (firmware upgrade)

rmorandell-pgum:
# Community contributors

## Description

- The unknown, critical & warning status is not se for the general API HTTP request. They are only set on login. So when token has a valid validation date in the statefile but is not more valid (because retired) the other checks doesn't terminate before validating their data.
- During a firmware upgrade the token saved in the statefile will not be valid anymore even it should by validation date. In this case the user must find the statefile an delete it to force a new login.

**Fixes** # (issue)
 With a simple clean of the token and a retry we can retry ones with a new login. On this retry we can set the http-status parameter so when the retry fails the check will be aborted with a status error.

## Type of change

### DIFF
--- a/src/storage/hp/primera/restapi/custom/api.pm
+++ b/src/storage/hp/primera/restapi/custom/api.pm
@@ -175,6 +175,21 @@ sub request_api {
         critical_status => ''
     );
 
+    # Maybe token is invalid. so we retry
+    if (!defined($self->{token}) && $self->{http}->get_code() < 200 || $self->{http}->get_code() >= 300) {
+        $self->clean_token();
+        $token = $self->get_token();
+
+        $content = $self->{http}->request(
+            url_path        => $options{endpoint},
+            get_param       => $get_param,
+            header          => [ 'X-HP3PAR-WSAPI-SessionKey: ' . $token ],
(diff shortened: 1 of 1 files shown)


src/storage/hp/primera/restapi/custom/api.pm
```diff
@@ -175,6 +175,21 @@ sub request_api {
         critical_status => ''
     );
 
+    # Maybe token is invalid. so we retry
```
omercier, line 178: This condition is equivalent to 
```
    if (
          ( !defined($self->{token}) && $self->{http}->get_code() < 200 )
          || $self->{http}->get_code() >= 300) {
```
which is probably not what you meant.

I suggest 

```
    if (!defined($self->{token}) || $self->{http}->get_code() >= 400) {
```
which stands for `if no token or if unauthorized`

rmorandell-pgum: yes you are right ! this is a typo .. the && should be a ||, too.

if no token or >= 400 is fine for me. thanks

omercier:
--> #5267 
Thanks again @rmorandell-pgum

PR #5262, opened 2024-10-24
Ctor 1019 plugin centreon plugin applications databases postgresql modes hitratio and backends exclude option is behaving the opposite way more like a filter

s-duret:
# Centreon team

## Description

Exclude options is actually including database in place of excluding them (4 modes concerned).
One mode has already been fixed [here](https://github.com/centreon/centreon-plugins/pull/4197).

**Fixes** # CTOR-1019

## Type of change

### DIFF
--- a/src/database/postgres/mode/listdatabases.pm
+++ b/src/database/postgres/mode/listdatabases.pm
@@ -53,7 +53,7 @@ sub manage_selection {
     );
     $self->{list_db} = [];
     while ((my $row = $self->{sql}->fetchrow_hashref())) {
-        if (defined($self->{option_results}->{exclude}) && $row->{datname} !~ /$self->{option_results}->{exclude}/) {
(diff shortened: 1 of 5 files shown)


omercier (approved):

src/database/postgres/mode/listdatabases.pm
```diff
@@ -1,4 +1,4 @@
-#
+    #
```
omercier, line 1: Why this?

lucie-dubrunfaut: I didn't add it :< (maybe a Pycharm parameter)

lucie-dubrunfaut, line 1:
```suggestion
#
```

omercier (approved):

PR #5296, opened 2024-11-21
CTOR 1176: Plugin(os::freebsd::snmp) - Mode(storage): add default ZFS support + tests

itoussies:
# Description

The latest version of pfSense (2.7.2) is based on freeBSD 14 and has ZFS as the default file system. This PR enhance the storage mode to support ZFS file system monitoring by default. Tests are also added.


tests/os/freebsd/snmp/list-diskspath.robot
```diff
@@ -0,0 +1,35 @@
...
+    ...    --snmp-version=${SNMPVERSION}
+    ...    --snmp-port=${SNMPPORT}
+    ...    --snmp-community=os/linux/snmp/linux
+    ...    --snmp-timeout=1
+    ...    ${extra_options}
+
+    Ctn Run Command And Check Result As Strings    ${command}    ${expected_result}
+
+    Examples:        tc    extra_options                           expected_result    --
+            ...      1     --skip-total-size-zero=''               List disk path: ${SPACE}'/run/user/0' [id = 25] ${SPACE} '/run' [id = 5] ${SPACE} '/' [id = 6] ${SPACE} '/dev/shm' [id = 8] ${SPACE} '/run/lock' [id = 9]
```
Evan-Adam, line 29: I tryied whithout the ${space} variable this file still work, why did you add it here ?

tests/os/freebsd/snmp/cpu-detailed.robot
```diff
@@ -0,0 +1,49 @@
...
+    ...    --snmp-version=${SNMPVERSION}
+    ...    --snmp-port=${SNMPPORT}
+    ...    --snmp-community=os/linux/snmp/linux
+    ...    --snmp-timeout=1
+    ...    ${extra_options}
+ 
+    Ctn Run Command And Check Result As Strings    ${command}    ${expected_result}
+
+    Examples:        tc    extra_options                   expected_result    --
+            ...      1     --critical-guest                OK: CPU Usage: user : Buffer creation, nice : Buffer creation, system : Buffer creation, idle : Buffer creation, wait : Buffer creation, kernel : Buffer creation, interrupt : Buffer creation, softirq : Buffer creation, steal : Buffer creation, guest : Buffer creation, guestnice : Buffer creation
```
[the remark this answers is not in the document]
Evan-Adam: Olivier added a new keyword Ctn Generic Suite Setup to remove the cache betweend each test to be sure the test are reproductible every try.
maybe it's because of the cache you have this output every time in this test.
(if it's true please add the new keyword in every file as it will become a best practice)

tests/os/freebsd/snmp/cpu-detailed.robot
```diff
@@ -0,0 +1,49 @@
+*** Settings ***
```
Evan-Adam, line 1: Please do not copy paste test from another if you use snmp_standard : in the end it's the same code, we need to test it only once, and it's harder to maintain as it is duplicated.
If you want to add more test you can make a test specific for you case, for exemple checking that without parameter the zfs filesystem is correctly monitored.

Evan-Adam (changes requested):

tests/os/freebsd/snmp/cpu-detailed.robot
itoussies: Yeah you're right, I removed tests using snmp_standard. I removed the _memory_ test too as the new machine I used (FreeBSD 14 on AWS) didn't have the needed OIDs. The previous walk was probably wrong. 
I kept the _storage_ test but I could not test the change on that machine as I can't create ZFS pools on AWS. The changes works on the client side however.

Evan-Adam (approved):

src/os/freebsd/snmp/mode/storage.pm
```diff
@@ -0,0 +1,132 @@
...
+
+=item B<--oid-display>
+
+Choose OID used to display storage (default: hrStorageDescr) (values: hrStorageDescr, hrFSMountPoint).
+
+=item B<--display-transform-src> B<--display-transform-dst>
+
+Modify the storage name displayed by using a regular expression.
+
+Example: adding --display-transform-src='dev' --display-transform-dst='run' will replace all occurrences of 'dev' with 'run'
```
omercier, line 110:
```suggestion
Example: adding C<--display-transform-src='dev' --display-transform-dst='run'> will replace all occurrences of C<dev> with C<run>.
```

omercier:
Other suggestions (could not do them in gh suggestions) :
- Freebsd --> FreeBSD in plugin.pm
- `Some modes ('cpu', 'load, 'swap', 'memory') needs 'bsnmp-ucd'.` --> `Some modes (C<cpu>, C<load>, C<swap>, C<memory>) need C<bsnmp-ucd>.`

src/os/freebsd/snmp/mode/storage.pm
itoussies: Done ✔️

omercier (approved):

Evan-Adam (approved):

PR #5564, opened 2025-04-28
Extend QNAP plugin for type QuTS hero

rmorandell-pgum:
# Community contributors

## Description

QNAP also has a QuTS Operating System which is used for certain components as well as for the pools and volumes on other OIDs.

Attached the MIB file

[QNAP_NAS.mib.txt](https://github.com/user-attachments/files/19939717/QNAP_NAS.mib.txt)

The snmpwalk i will deliver via email.

In the MIB file there should be also the OIDs for the msatadisk infos but in my case my devices doesn't respond on this OID's so i don't extend the mdisk check yet.

## Type of change

### DIFF
--- a/src/storage/qnap/snmp/mode/components/disk.pm
+++ b/src/storage/qnap/snmp/mode/components/disk.pm
@@ -24,41 +24,48 @@ use strict;
 use warnings;
 
 my $map_status_disk = {
-    0 => 'ready',
+    0    => 'ready',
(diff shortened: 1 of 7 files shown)


lucie-tirand:
Hello :)

This PR is handle in the June release see [here](https://docs.centreon.com/fr/pp/integrations/plugin-packs/releases/release-notes/).
Thanks again for your contribution :)

PR #5610, opened 2025-06-04
Plugin(database::mysql::plugin) - Mode(dbi,mysqlcmd): Error when retrieving the server version

itoussies:
Refs:CTOR-1562

# Centreon team (internal PR)

## Description

The regex doesn't work for strings like '11.4.4-2-MariaDB-enterprise-log'. This PR adapts the regex to handle this case. 
The regex seems more complicated, `([0-9\.\-]*?)-MariaDB` would be sufficient but it doesn't return the same result as before. For example for `5.5.5-10.1.36-MariaDB`, the group 1 would be `5.5.5-10.1.36` instead of `10.1.36`. It's working fine so you tell me which one you prefer.

## Type of change

- [X] Functionality enhancement or optimization (non-breaking change)

### DIFF
--- a/src/database/mysql/mysqlcmd.pm
+++ b/src/database/mysql/mysqlcmd.pm
@@ -209,8 +209,8 @@ sub set_version {
 
     $self->{is_mariadb} = 0;
     $self->{version} = $options{version};
-    # MariaDB: 5.5.5-10.1.36-MariaDB or 10.1.36-MariaDB
-    if ($self->{version} =~ /([0-9\.]*?)-MariaDB/i) {
+    # MariaDB: 5.5.5-10.1.36-MariaDB, 10.1.36-MariaDB or 11.4.4-2-MariaDB-enterprise-log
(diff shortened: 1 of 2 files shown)


src/database/mysql/mysqlcmd.pm
```diff
@@ -307,13 +307,13 @@ __END__
...
-mysqlcmd global
+MySQL command global
 
 =head1 SYNOPSIS
 
-mysqlcmd class
+MySQL command class
 
-=head1 MYSQLCMD OPTIONS
+=head1 MYSQL COMMAND OPTIONS
```
omercier, line 316: This change will beak the help option (l. 56) 
```perl
$options{options}->add_help(package => __PACKAGE__, sections => 'MYSQLCMD OPTIONS', once => 1);
```

itoussies: Fixed, thanks 👍🏾

PR #5634, opened 2025-06-19
CTOR-1248: Display full curl command in log

scresto31:
# Centreon team (internal PR)

## Description

Display the curl command if --debug option is set

**Fixes** # CTOR-1248

## Type of change


kduret (approved):

tuntoja (approved):

src/centreon/plugins/backend/http/curl.pm
```diff
@@ -158,23 +160,39 @@ sub set_method {
...
 
     if ($options{request}->{method} eq 'GET') {
-        return ;
+        # no curl_log call because GET is the default value
+        return;
     }
 
     if ($options{request}->{method} eq 'POST') {
         $self->curl_setopt(option => $self->{constant_cb}->(name => 'CURLOPT_POST'), parameter => 1);
+        $self->{curl_log}->log('-X', $options{request}->{method}) unless $skip_log_post;
```
Evan-Adam, line 191: is it normal to log twice if the method is POST ?
I think you can remove this call to log() ?
(I think the return should be removed too but i'm not sure)

src/centreon/plugins/curllogger.pm
```diff
@@ -0,0 +1,121 @@
...
+# Conversion of some parameters manually passed to the set_extra_curl_opt function
+# into their command-line equivalents. Only the parameters used in the plugin code
+# are handled. If new parameters are added, this hash must be updated.
+# The hash contains curl parameters CURLOPT_*. Its keys map to either a hash when
+# multiple values are handled, or directly to an array when only one response is
+# supported. The placeholder <value> is replaced by the provided value.
+# Eg: "CURLOPT_POSTREDIR => CURL_REDIR_POST_ALL" will produce: --post301 --post302 --post303
+#     "CURLOPT_SSL_VERIFYPEER => 0" will produce: --insecure
+#     "CURLOPT_AWS_SIGV4 => 'osc'" will produce: --aws-sigv4 osc
+our %curlopt_to_parameter = (
```
Evan-Adam, line 82: can you manage CURLOPT_SSLVERSION ?
(it's in the curl.pm documentation)

src/centreon/plugins/backend/http/curl.pm
```diff
@@ -261,14 +296,18 @@ sub set_extra_curl_opt {
...
         if (!defined($entries->{$key})) {
-            $entries->{$key} = { val => [], force_array => 0 };
+            $entries->{$key} = { val => [], constants => [], force_array => 0 };
         }
 
         $value = centreon::plugins::misc::trim($value);
         if ($value =~ /^\[(.*)\]$/) {
             $entries->{$key}->{force_array} = 1;
             $value = centreon::plugins::misc::trim($1);
+            push @{$entries->{$key}->{constants}}, map { $_ = centreon::plugins::misc::trim($_); s/^'(.*)'$/$1/; $_  } split ',', $value;
```
Evan-Adam, line 323: can you put a comment to explain what you do here please ?

sdepassio (approved):

src/centreon/plugins/backend/http/curl.pm
scresto31: The goal of this return is to avoid adding a redundant -X POST to the curl command. The POST method is sometimes implicit, for example when using the --data-raw option curl automatically switches to POST. Therefore there's no need to generate a command like curl -X POST --data-raw blabla http://localhost, since curl --data-raw blabla http://localhost does the same

src/centreon/plugins/backend/http/curl.pm
scresto31: Ok, I will add comment.
The goal is to convert a string containing curl options into an array, removing the quotes, for example:

`my $opt = ["CURLOPT_SSL_VERIFYPEER =>[opt1,'opt2','opt3']"];
`

changes to:

```
$VAR1 = [
  'opt1',
  'opt2',
  'opt3'
];
```

sdepassio (dismissed):

Evan-Adam (dismissed):

Evan-Adam (approved):

tuntoja (approved):

PR #5645, opened 2025-06-27
CTOR-1558: Enhance standby mode and extra parameters support

scresto31:
## Description

Plugin(apps::hashicorp::vault::restapi) - Mode(health)

Enhance standby mode support and add support for extra parameters

**Fixes** # CTOR-1558

## Type of change

- [x] Functionality enhancement or optimization (non-breaking change)

### DIFF
--- a/src/apps/hashicorp/vault/restapi/mode/health.pm
+++ b/src/apps/hashicorp/vault/restapi/mode/health.pm
@@ -47,6 +47,13 @@ sub set_counters {
                 closure_custom_perfdata => sub { return 0; },
                 closure_custom_threshold_check => \&catalog_status_threshold_ng
             }
+        },
+        { label => 'standby-status', type => 2, set => {
+                key_values => [ { name => 'standby' } ],
+                output_template => "standby status : %s",
+                closure_custom_perfdata => sub { return 0; },
+                closure_custom_threshold_check => \&catalog_status_threshold_ng
+            }
         }
     ];
 }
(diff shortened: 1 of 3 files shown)


src/apps/hashicorp/vault/restapi/mode/health.pm
```diff
@@ -71,17 +98,52 @@ sub set_options {
     my ($self, %options) = @_;
 }
 
+sub check_options {
+    my ($self, %options) = @_;
+    $self->SUPER::check_options(%options);
+
+    foreach my $param (@code_parameters) {
+	next unless $param->{'type'};
```
Evan-Adam: Indentation seem wrong here

Evan-Adam: Why do you check if type exist here ?
the array seem defined entirely in the code ? can it be updated by the user ?

Evan-Adam: (same question for $value)

src/apps/hashicorp/vault/restapi/mode/health.pm
```diff
@@ -71,17 +98,52 @@ sub set_options {
...
+	my $value = lc $self->{option_results}->{$param->{'code'}};
+	next if $value eq '';
+
+	my $valid = 0;
+	if ($param->{'type'} eq 'status') {
+	    $valid = $value =~ /^\d{1,3}$/;
+	} elsif ($param->{'type'} eq 'bool') {
+	    $valid = $value eq 'true' || $value eq 'false';
```
Evan-Adam: optional : make the lc before the condition to silently fix and ignore when a user make a mistake

src/apps/hashicorp/vault/restapi/mode/health.pm
scresto31: > (same question for $value)

$value comes from the command line, we ignore empty command-line parameters

src/apps/hashicorp/vault/restapi/mode/health.pm
scresto31: Yes, that's it (line 106), and I removed the unnecessary lc at the end

PR #5723, opened 2025-08-27
CTOR-672 Plugin(os::linux::local) - Mode(ntp) : Adding support for timedatectl

scresto31:
## Description

Plugin(os::linux::local) - Mode(ntp) : Adding support for timedatectl

**Fixes** # CTOR-672

## Type of change


src/os/linux/local/mode/ntp.pm
```diff
@@ -68,6 +68,28 @@ my %unit_map_chronyc = (
...
+    'h' => 60 * 60 * 1000,
+    'd' => 24 * 60 *60 * 1000
+);
+
+my %state_map_timedatectl = (
+    'synchronized' => 'currently active and fully synchronized',
+    'syncing' => 'currently active and synchronizing',
+    'available' => 'configured and available for use',
+    'unused' => 'configured as fallback but not used',
+    'inactive' => 'not used besauve NTP service is inactive'
```
omercier: `because`?

src/os/linux/local/mode/ntp.pm
```diff
@@ -209,39 +232,166 @@ sub get_ntp_modes {
...
+    # timedatectl show-timesync to retrieve 'SystemNTPServers', 'FallbackNTPServers', 'ServerAddress' and 'ServerName'
+    my ($stdout_status) = $options{custom}->execute_command(    command => 'timedatectl',
+                                                                command_options => 'status 2>&1',
+                                                                no_quit => 1, );
+    my ($stdout_timesync) = $options{custom}->execute_command(  command => 'timedatectl',
+                                                                command_options => 'timesync-status 2>&1',
+                                                                no_quit => 1, );
+    my ($stdout_show) = $options{custom}->execute_command(      command => 'timedatectl',
+                                                                command_options => 'show-timesync 2>&1',
+                                                                no_quit => 1, );
```
omercier, line 315: A sample output of each command would be welcome as comments to ease code comprehension

src/os/linux/local/mode/ntp.pm
```diff
@@ -209,39 +232,166 @@ sub get_ntp_modes {
...
+
+    $values{$_}//='' foreach ('SystemNTPServers', 'FallbackNTPServers', 'ServerAddress', 'ServerName', 'Offset', 'Stratum', 'Packet count', 'System clock synchronized');
+
+    my $active_is_fallback = 0;
+
+    # Primary and fallback servers are initialized, the active server is excluded as it will be initialized later with additional information
+    # A server has either the type 'primary' or 'fallback'
+    # A 'primary' server can have the states 'available', 'synchronized', 'syncing' or 'inactive'
+    # A 'fallback' server can have the states 'unused', synchronized', 'syncing' or 'inactive'
+    foreach (split /\s/, $values{SystemNTPServers}) {
```
omercier: I'm not a fan of implicit $_ variable when it can be explicited. Why not `foreach my $server (split /\s/, $values{SystemNTPServers}) {`

src/os/linux/local/mode/ntp.pm
```diff
@@ -209,39 +232,166 @@ sub get_ntp_modes {
...
+        next if $values{ServerAddress} && ($_ eq $values{ServerAddress} || $_ eq $values{ServerName});
```
omercier: I don't get why this case must be skipped. An explanation in comment would help

src/os/linux/local/mode/ntp.pm
```diff
@@ -209,39 +232,166 @@ sub get_ntp_modes {
...
+
+        $self->{peers}->{$_} = { display => $_, rawstate => 'available', stratum => '', rawtype => 'primary', reach => '', offset => '' };
+        $self->{global}->{peers}++
+    }
+    foreach (split /\s/, $values{FallbackNTPServers}) {
```
omercier: Same remark about `$_`

src/os/linux/local/mode/ntp.pm
```diff
@@ -209,39 +232,166 @@ sub get_ntp_modes {
...
+            next
+        }
+        $self->{peers}->{$_} = { display => $_, rawstate => 'unused', stratum => '', rawtype => 'fallback', reach => '', offset => '' };
+        $self->{global}->{peers}++
+    }
+
+    if ($values{ServerAddress} ne '') {
+        # If there is an active server it is initialized here with all information
+        $values{ServerName} = $values{ServerAddress} if $values{ServerName} eq '';
+        $values{'Offset'} = $1 * $unit_map_timedatectl{$2} if $values{'Offset'} =~ /^(.*?)([a-z]+)$/ && $unit_map_timedatectl{$2};
```
omercier, line 349:
```suggestion
        $values{'Offset'} = $1 * $unit_map_timedatectl{$2} if ( $values{'Offset'} =~ /^(.*?)([a-z]+)$/ && defined($unit_map_timedatectl{$2}) );
```

omercier: I know that it will result in the same behaviour given none of the hash's keys is equal to 0 but I think it is a better practice to test it this way since a value of 0 would give false in this condition

src/os/linux/local/mode/ntp.pm
```diff
@@ -209,39 +232,166 @@ sub get_ntp_modes {
...
     };
 
-    if ($self->{option_results}->{ntp_mode} eq 'ntpq') {
-        return [ $modes->{ntpq} ];
-    } elsif ($self->{option_results}->{ntp_mode} eq 'chronyc') {
-        return [ $modes->{chronyc} ];
+    # Returns the selected mode or all modes
+    return [ $modes->{$self->{option_results}->{ntp_mode}} ?
+                $modes->{$self->{option_results}->{ntp_mode}} :
+                ( $modes->{timedatectl}, $modes->{ntpq}, $modes->{chronyc} ) ];
```
omercier:
```suggestion
                ( $modes->{timedatectl}, $modes->{chronyc}, $modes->{ntpq} ) ];
```

omercier: I suggest this because ntpq is older than chrony, which is older than timedatectl, and older should also mean rarer

src/os/linux/local/mode/ntp.pm
```diff
@@ -209,39 +232,166 @@ sub get_ntp_modes {
...
+                                                     rawstate => $values{'System clock synchronized'} eq 'yes' ? 'synchronized' : 'syncing',
+                                                     rawtype => $active_is_fallback ? 'fallback' : 'primary',
+                                                     stratum => $values{Stratum},
+                                                     reach => $values{'Packet count'},
+                                                     offset => $values{'Offset'} };
+        $self->{global}->{peers}++;
+    }
+
+    foreach (keys %{$self->{peers}}) {
+        $self->{peers}->{$_}->{rawstate} = 'inactive' if $values{'NTP service'} eq 'inactive';
```
omercier: At this point I don't see how a primary server could ever still be in `available` rawstate.
Am I right?

src/os/linux/local/mode/ntp.pm
scresto31: The explanation is in the comment a few lines above: "the active server is excluded as it  will be initialized later with additional information"

src/os/linux/local/mode/ntp.pm
scresto31: I’m not sure I agree here, the extra parentheses and the unnecessary `defined` just make the line harder to read, since unit_map_timedatectl never returns 0

src/os/linux/local/mode/ntp.pm
scresto31: The primary servers are initialized to `available` when the hash is created,
and only one of the active servers may have been changed to `synchronized` or `syncing`

omercier (approved):

PR #5732, opened 2025-09-03
CTOR-1409 Plugin(network-security-cato-networks-api) - Mode(discovery)

scresto31:
## Description

Plugin(network-security-cato-networks-api) - Mode(discovery)

**Fixes** # CTOR-1409

## Type of change

### DIFF
--- a/src/centreon/plugins/misc.pm
+++ b/src/centreon/plugins/misc.pm
@@ -24,6 +24,9 @@ use strict;
 use warnings;
 use utf8;
 use JSON::XS;
+use Exporter 'import';
+
+our @EXPORT_OK = qw/value_of json_encode json_decode/;
 
 sub execute {
     my (%options) = @_;
@@ -338,6 +341,20 @@ sub is_empty {
         return 1;
     }
     return 0;
+}
+
+# Return the value of a complex perl variable (hash, array...) or a default value if it not defined.
+# The returned value will never be undef.
+# I.g:  value_of($hash, '->{key}->{subkey}', 'default')
+#       value_of($array, '->[0]', 'default')
+#       value_of($complex, '->{key}->[0]->{subkey}', 'default')
+sub value_of($$;$) {
+    my ($variable, $expression, $default) = @_;
+    $default //= '';
+
+    my $value = eval "\$variable$expression";
+
+    return defined $value ? $value : $default;
 }
 
 sub trim {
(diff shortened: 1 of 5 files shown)


src/centreon/plugins/misc.pm
```diff
@@ -340,6 +343,20 @@ sub is_empty {
     return 0;
 }
 
+# Return the value of a complex perl variable (hash, array...) or a default value if it not defined.
+# The returned value will never be undef.
+# I.g:  value_of($hash, '->{key}->{subkey}', 'default')
+#       value_of($array, '->[0]', 'default')
+#       value_of($complex, '->{key}->[0]->{subkey}', 'default')
+sub value_of($$;$) {
```
Evan-Adam, line 352: I understand that for now the second param is hardcoded in the code.
However, this may not be the case in the future, and could allow remote code execution if an attacker take control of this variable.
Can you check if you can use the Safe lib for this ?
maybe you can move this function in an object to avoid creating the safe object each time ?
I see the object is already setup in output.pm

src/network/security/cato/networks/api/custom/api.pm
```diff
@@ -0,0 +1,321 @@
...
+    $self->{http} = centreon::plugins::http->new(%options, default_backend => 'curl');
+    $self->{cache} = centreon::plugins::statefile->new(%options);
+
+    return $self;
+}
+
+sub check_options {
+    my ($self, %options) = @_;
+
+    foreach my $opt (qw/account_id api_key port hostname endpoint proto max_retry_count retry_delay timeout/) {
```
Evan-Adam, line 71: I think you can remove check of option that have a default value no ?

src/network/security/cato/networks/api/custom/api.pm
```diff
@@ -0,0 +1,321 @@
...
+    }
+}
+
+sub set_options {
+    my ($self, %options) = @_;
+
+    $self->{option_results} = $options{option_results};
+}
+
+sub set_defaults {}
```
Evan-Adam, line 87: Why is this necessary ?
I don't see any base() so it's not to override a default ?

src/network/security/cato/networks/api/mode/discovery.pm
```diff
@@ -0,0 +1,126 @@
...
+    $self->{$_} = $self->{option_results}->{$_} for qw/filter_site_name connectivity_details/;
+
+    $self->{filter_site_id} = [ map { split /,/ } @{$self->{option_results}->{filter_site_id}} ];
+
+}
+
+sub manage_selection {
+    my ($self, %options) = @_;
+
+    #my $results = undef;
```
Evan-Adam: To remove ?

src/network/security/cato/networks/api/custom/api.pm
scresto31: The check is necessary because users can explicitly set these values to empty strings on command line ( --hostname='' )

src/network/security/cato/networks/api/custom/api.pm
scresto31: `set_options` and `set_defaults` are necessary because they are called by script_custom.pm at lines 153/154

src/centreon/plugins/misc.pm
scresto31: OK, I’m now using `Safe` and state to create the object only once

PR #5777, opened 2025-09-29
CTOR-111 Plugin(apps::monitoring::splunk) - Mode(query): extend this mode

scresto31:
## Description

Plugin(apps::monitoring::splunk) - Mode(query): extend this mode

**Fixes** # CTOR-111

## Type of change

### DIFF
--- a/packaging/centreon-plugin-Applications-Monitoring-Splunk-Api/deb.json
+++ b/packaging/centreon-plugin-Applications-Monitoring-Splunk-Api/deb.json
@@ -1,6 +1,6 @@
 {
     "dependencies": [
         "libdatetime-perl",
-	"libxml-simple-perl"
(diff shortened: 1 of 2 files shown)


src/apps/monitoring/splunk/custom/api.pm
```diff
@@ -117,6 +119,21 @@ sub build_options_for_httplib {
     $self->{option_results}->{proto} = $self->{proto};
 }
 
+sub json_decode {
```
sdepassio, line 122: There is a "json_decode" in misc.pm, I think you can use it instead of this new function.

scresto31: Yes, I chose to not use it because it outputs its error messages to stderr

PR #5780, opened 2025-10-01
Update dcdiag.pm - add DE language

joschi99:
This is the same language setting for german used on the "local" command.


lucie-tirand:
Replaced by https://github.com/centreon/centreon-plugins/pull/5844

Thanks again for your contribution :)

PR #5791, opened 2025-10-02
CTOR-1871 Plugin(storage::ibm::storwize::ssh) - Custom(api): issue while parsing response

scresto31:
# Community contributors

## Description

**Fixes** # CTOR-1871
Plugin(storage::ibm::storwize::ssh) - Custom(api): issue while parsing response

## Type of change

### DIFF
--- a/src/storage/ibm/storwize/ssh/custom/api.pm
+++ b/src/storage/ibm/storwize/ssh/custom/api.pm
@@ -87,16 +87,36 @@ sub get_hasharray {
 
     my $result = [];
     return $result if ($options{content} eq '');
-    my ($header, @lines) = split /\n/, $options{content};
-    my @header_names = split /$options{delim}/, $header;
-    
-    for (my $i = 0; $i <= $#lines; $i++) {
-        my @content = split /$options{delim}/, $lines[$i];
-        my $data = {};
-        for (my $j = 0; $j <= $#header_names; $j++) {
-            $data->{$header_names[$j]} = $content[$j];
+
+    my @lines = split /\n/, $options{content};
+    my $header;
+    my @clean_lines;
+
+    while (@lines) {
+        my $line = shift @lines;
+        my @fields = split /$options{delim}/, $line;
+
+        # Check if this line is the header
+        if (scalar(@fields) > 1 ) {
+            # Found the header — keep this line and the rest
+            $header = $line;
+            @clean_lines = @lines;
+            last;
         }
-        push @$result, $data;
+        # skip line and continue
+    }
+
+    if (@clean_lines) {
+        my @header_names = split /$options{delim}/, $header;
+
+        for (my $i = 0; $i <= $#clean_lines; $i++) {
+            my @content = split /$options{delim}/, $clean_lines[$i];
+            my $data = {};
+            for (my $j = 0; $j <= $#header_names; $j++) {
+                $data->{$header_names[$j]} = $content[$j];
+            }
+            push @$result, $data;
+        }
     }
 
     return $result;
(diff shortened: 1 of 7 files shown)


src/storage/ibm/storwize/ssh/custom/api.pm
```diff
@@ -107,14 +111,30 @@ sub execute_command {
...
+            $command_prefix = qq(echo '$prefix';);
+            $command_suffix = qq(;echo '$suffix';);
+        }
+        $command = defined($self->{option_results}->{command}) && $self->{option_results}->{command} ne '' ? $self->{option_results}->{command} : $options{command};
         ($content) = $self->{ssh}->execute(
             hostname => $self->{option_results}->{hostname},
-            command => defined($self->{option_results}->{command}) && $self->{option_results}->{command} ne '' ? $self->{option_results}->{command} : $options{command},
+            command => $command_prefix.$command.$command_suffix,
             command_path => $self->{option_results}->{command_path},
             command_options => defined($self->{option_results}->{command_options}) && $self->{option_results}->{command_options} ne '' ? $self->{option_results}->{command_options} : undef,
```
Evan-Adam, line 127: I think you should put the command_suffix after this no ?
because if client set the argument of the command in command option the command will be wrong ?
likely because of this : src/centreon/plugins/backend/ssh/libssh.pm:111

Same for command prefix and command_path, i think it's concatened by the lib

Maybe you can change only the sshcli backend ?
if you reproduced it do you reproduce it with all backend ?

Evan-Adam: I know there is also the -q (quiet) of ssh to ignore some error message (but it's a pretty "recent" option)

tests/storage/ibm/storwize/ssh/api.t
```diff
@@ -0,0 +1,115 @@
...
+        print $fic "#!/bin/sh\n";
+        if ($test->{banner}) {
+            print $fic qq(cat<<EOF\n).
+                       $test->{banner}.
+                       qq(EOF\n);
+        }
+        close($fic);
+
+        # Test command execution
+        $result = $api->execute_command(wrap_command => 1);
```
Evan-Adam, line 104: you should probably add test for command_path and command_options of this sub

tests/storage/ibm/storwize/ssh/mode.robot
```diff
@@ -0,0 +1,32 @@
...
+    [Tags]    storage    ibm    storwize    ssh
+    ${command}    Catenate
+    ...    ${CMD}
+    ...    ${extra_options}
+
+    Ctn Run Command And Check Result As Strings    ${command}    ${expected_string}
+
+    Examples:      tc       extra_options             expected_string     --
+        ...        1        ${EMPTY}                  UNKNOWN: Need to specify '--mode' or '--dyn-mode' option.
+        ...        2        --mode=eventlog           UNKNOWN: please set --hostname option for ssh connection (or --command for local)
```
Evan-Adam, line 29: Plugin pack tests check a mode exist, so I don't think it's usefull to check the mode existence here
Maybe you can make only one test (by using the command override) to check a real output for only one mode instead ?

Evan-Adam (changes requested):

src/storage/ibm/storwize/ssh/custom/api.pm
scresto31: Using an SSH option seems fine, I've opened a new PR for that

scresto31:
This PR is replaced by https://github.com/centreon/centreon-plugins/pull/5800

PR #5816, opened 2025-10-17
CTOR-1919 Plugin(cloud::openstack::restapi) - Mode (discovery), Mode(list-services), Mode(service)

scresto31:
## Description

Plugin(cloud::openstack::restapi) - Mode (discovery), Mode(list-services), Mode(service)

**Fixes** # CTOR-1919 

## Type of change

### DIFF
--- a/src/cloud/openstack/restapi/custom/api.pm
+++ b/src/cloud/openstack/restapi/custom/api.pm
@@ -62,15 +62,15 @@ sub new {
 
         _services_options(name=>'keystone', port => 5000, endpoint => '/v3/auth/tokens'),
 
-        'username:s'        => { name => 'username', default => '' },
-        'password:s'        => { name => 'password', default => '' },
-        'user-domain:s'     => { name => 'user_domain', default => 'default' },
-        'project-name:s'    => { name => 'project_name', default => 'admin' },
-        'project_domain:s'  => { name => 'project_domain', default => 'default' },
-        'authent-by-env:s'  => { name => 'authent_by_env', default => '0' },
-        'authent-by-file:s' => { name => 'authent_by_file', default => '' },
+        'username:s'          => { name => 'username', default => '' },
+        'password:s'          => { name => 'password', default => '' },
+        'user-domain-id:s'    => { name => 'user_domain_id', default => 'default' },
+        'project-name:s'      => { name => 'project_name', default => 'admin' },
+        'project-domain-id:s' => { name => 'project_domain_id', default => 'default' },
+        'authent-by-env:s'    => { name => 'authent_by_env', default => '0' },
+        'authent-by-file:s'   => { name => 'authent_by_file', default => '' },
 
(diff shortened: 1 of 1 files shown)


src/cloud/openstack/restapi/custom/api.pm
```diff
@@ -0,0 +1,437 @@
...
+        }
+    }
+
+    if ($options{apply_conf_from_file}) {
+        open(my $file, "<".$options{apply_conf_from_file})
+            or $self->{output}->option_exit(short_msg => "Cannot open file '".$options{apply_conf_from_file}."': $!");
+        foreach my $line (<$file>) {
+            next unless $line =~ /^[\s\t]*export[\t\s]+(\w+)=["']?(.*?)["']?$/;
+
+            # Only handle variables defined using 'export variable="value"'
```
omercier, line 115: So the config file only aims at overloading values already available in the environment? Can you add a link to the documentation in the comments?

src/cloud/openstack/restapi/mode/discovery.pm
```diff
@@ -0,0 +1,144 @@
...
+                my $horizon_url = $endpoint->{url} =~ s/^(https?:\/\/[-\.\w]+).*$/$1/r;
+                next if $test{$horizon_url};
+                $test{$horizon_url} = 1;
+                my $result = $options{custom}->ping_service(type => 'dashboard',
+                                                            service_url => $horizon_url,
+                                                            expected_data => 'auto',
+                                                            endpoint_suffix => 'auto',
+                                                            token => '');
+                push @horizon_url, $horizon_url
+                    if $result->{http_status} == 200 && $result->{valid_content};
```
omercier, line 82: Should we not set `$test{$horizon_url} = 1;` after the ping has revealed successful?

src/cloud/openstack/restapi/custom/api.pm
scresto31: Ok, I added a link, and I also added the missing _ID to the variable names

src/cloud/openstack/restapi/mode/discovery.pm
scresto31: No this is intentional because I only want to test each URL once since a second attempt wouldn't make it succeed if it fails the first time

omercier (approved):

src/cloud/openstack/restapi/custom/api.pm
```diff
@@ -0,0 +1,439 @@
...
+use centreon::plugins::misc qw/json_encode json_decode value_of/;
+use Digest::MD5 qw(md5_hex);
+use DateTime::Format::Strptime;
+
+# Define all specific options for a service
+sub _services_options {
+    my (%options) = @_;
+    my $name = $options{name};
+
+    ( $name.'-url:s'      => { name => $name.'_url', default => '' },
```
Evan-Adam: Please add return and a ; to make clear what you want to return

PR #5817, opened 2025-10-17
CTOR-541 New Pack(cloud-openstack-restapi)

scresto31:
## Description

New Pack(cloud-openstack-restapi)

**Fixes** # CTOR-541

## Type of change


garnier-quentin:
@scresto31 you don't need to do that:
```
'authent-by-env:s' => { name => 'authent_by_env', default => '0' },
```

There is a system inside options.pm to manage it (use following code. it works for all options of that type):
```
'authent-by-env' => { name => 'authent_by_env' },
...
if (defined($self->{option_results}->{authent_by_env})) {

}
```

And you can use option in command line:
- ```--authent-by-env``` or ```authent-by-env=1``` to enable it
- ```--authent-by-env=0``` to disable it

PR #5827, opened 2025-10-24
CTOR-1913 Plugin(cloud::openstack::restapi) - Mode(volume)

scresto31:
# Centreon team (internal PR)

## Description

Plugin(cloud::openstack::restapi) - Mode(volume)

**Fixes** # CTOR-1913

## Type of change

### DIFF
--- a/src/cloud/openstack/restapi/custom/api.pm
+++ b/src/cloud/openstack/restapi/custom/api.pm
@@ -89,7 +89,7 @@ sub new {
         _service_filters_options(type => 'compute', ),
         _service_ident_options(type => 'image',    port => 9292, endpoint => ''),
         _service_filters_options(type => 'image', ),
-        _service_ident_options(type => 'volume',   port => 9292, endpoint => ''),
+        _service_ident_options(type => 'volume',   port => 8776, endpoint => ''),
         _service_filters_options(type => 'volume', ),
 
         # Endpoint filters common to all services
@@ -935,7 +935,7 @@ Set the protocol to use in the Glance service URL (default: 'https').
 
 =item B<--image-port>
 
-Set the port to use in the Glance service URL (default: 8774).
+Set the port to use in the Glance service URL (default: 9292).
 
 =item B<--image-endpoint>
 
@@ -946,38 +946,38 @@ Set the endpoint to use in the Glance service URL (default: '/v2').
 Allow insecure TLS connection (default: '0').
 When set to 0 the default insecure value passed with --insecure is used.
 
-=item B<--image-url>
+=item B<--volume-url>
 
-Set the URL to use for the OpenStack Glance (image) service.
+Set the URL to use for the OpenStack Cinder (volume) service.
 A valid Glance URL is required since it is a mandatory service.
 Glance URL is retrieved from Keystone catalog unless disco-mode is set to 'manual' or a specific URL is provided with this option.
 
-Example: C<--image-url="https://myopenstack.local:9292">
+Example: C<--volume-url="https://myopenstack.local:8776">
 
-This URL can also be construct with options (--image-hostname, --image-proto, --image-port, --image-endpoint).
+This URL can also be construct with options (--volume-hostname, --volume-proto, --image-port, --image-endpoint).
 
-=item B<--image-hostname>
+=item B<--volume-hostname>
 
-Set the hostname part of the Glance service URL.
+Set the hostname part of the Cinder service URL.
 
-=item B<--image-proto>
+=item B<--volume-proto>
 
-Set the protocol to use in the Glance service URL (default: 'https').
+Set the protocol to use in the Cinder service URL (default: 'https').
 
-=item B<--image-port>
+=item B<--volume-port>
 
-Set the port to use in the Glance service URL (default: 8774).
+Set the port to use in the Cinder service URL (default: 8776).
 
-=item B<--image-endpoint>
+=item B<--volume-endpoint>
 
-Set the endpoint to use in the Glance service URL (default: '/v2').
+Set the endpoint to use in the Cinder service URL.
 
-=item B<--image-insecure>
+=item B<--volume-insecure>
 
 Allow insecure TLS connection (default: '0').
 When set to 0 the default insecure value passed with --insecure is used.
 
-=item B<--image-timeout>
+=item B<--volume-timeout>
 
 Set HTTP timeout in seconds (default: '0').
 When set to 0 the default timeout value passed with --timeout is used.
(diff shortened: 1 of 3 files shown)


src/cloud/openstack/restapi/custom/api.pm
```diff
@@ -338,6 +339,85 @@ sub connect_info {
...
+            $self->connect_info(url => $self->{volume_url}, resource => '/volumes/detail'),
+            insecure => $self->{volume_insecure},
+            critical_status => '',
+            warning_status => '',
+            unknown_status => ''
+        );
+
+        my $response = json_decode($response_brut);
+
+        return { http_status => $self->{http}->get_code(),message => value_of($response, "->{error}->{title}", 'Bad request').': '.value_of($response, "->{error}->{message}", "Invalid response") }
```
Evan-Adam: Please indent the complex oneliner, for exemple like this : 
```suggestion
    if (ref $response ne 'HASH' || $response->{error} || not $response->{volumes}){
        return { http_status => $self->{http}->get_code(),
            message =>
                value_of($response, "->{error}->{title}", 'Bad request')
                .': '
                .value_of($response, "->{error}->{message}", "Invalid response") };
    }

```

src/cloud/openstack/restapi/custom/api.pm
```diff
@@ -338,6 +339,85 @@ sub connect_info {
...
+        $params{$_} = $1 if $options{"include_".$_} =~ /^([\w\s]+)$/
+    }
+    $params{availability_zone} = $1 if $options{"include_zone"} =~ /^([\w\s]+)$/;
+
+    $params{project_id} = $options{project_id} if $options{project_id};
+
+    my $response_brut;
+    my @results;
+
+    # Retry to handle token expiration
```
Evan-Adam: Seem to me you made a loop to manage the pagination, not the authentication cache which seem managed in the volume.pm?

tests/cloud/openstack/restapi/volume.robot
```diff
@@ -0,0 +1,48 @@
...
+            ...      4     --include-name=vol-a-amoi-2                              OK: Volume count: 1 - Volume vol-a-amoi-2 is in available state, Id: ca382d6d-53a7-45e0-97d8-fec68abbdcc0, Type: __DEFAULT__, Description: un volume non admin 2, Size: 1, Project_id: dd2cef13c210457d9cdf06e24c2c37b0, Bootable: 0, Encrypted: 0, Zone: nova, Attachments count: 0 | 'volume.count'=1;;;0;
+            ...      5     --include-id=b510453a-6727-4392-947c-7431aa31250e        OK: Volume count: 1 - Volume vol-a-amoi is in available state, Id: b510453a-6727-4392-947c-7431aa31250e, Type: __DEFAULT__, Description: un volume non admin, Size: 1, Project_id: dd2cef13c210457d9cdf06e24c2c37b0, Bootable: 0, Encrypted: 0, Zone: nova, Attachments count: 0 | 'volume.count'=1;;;0;
+            ...      6     --exclude-status=available                               OK: Volume count: 0 | 'volume.count'=0;;;0;
```
Evan-Adam: I think if we don't find any element we should return unknown, but it can vary on the real use case, are you sure we should return OK when we can't found any volumes ?

src/cloud/openstack/restapi/mode/volume.pm
```diff
@@ -0,0 +1,377 @@
...
+Define the conditions to match for the status to be CRITICAL based on the volume size.
+Example: --critical-size='%{size} =~ /1 GB/'
+
+=item B<--warning-status>
+
+Define the conditions to match for the status to be WARNING based on the volume status.
+Example: --warning-statu='%{status} =~ /downloading/'
+Please refer to https://docs.openstack.org/api-ref/block-storage/v3/#volumes-status for more information about status.
+
+=item B<--critical-status>
```
Evan-Adam, line 351: I think you can add default value for this, for exemple error|error_deleting|error_backing-up|error_restoring|error_extending

src/cloud/openstack/restapi/mode/volume.pm
```diff
@@ -0,0 +1,377 @@
...
+
+Exclude by volume encrypted flag (can be 0 or 1).
+
+=item B<--include-zone>
+
+Filter by volume availability zone (can be a regexp and can be used multiple times or for comma separated values).
+
+=item B<--exclude-zone>
+
+Exclude by volume availability zone (can be a regexp and can be used multiple times or for comma separated values).
```
Evan-Adam, line 250: If the user want to match a regexp containing a comma, can they  or should they put a dot instead ?
Maybe we should not allows multiples value in the same value if it's a regex, as user can already mimick it with 'value1|value2' ?

src/cloud/openstack/restapi/mode/volume.pm
```diff
@@ -0,0 +1,377 @@
...
+    }
+}
+
+1;
+
+__END__
+
+=head1 MODE
+
+List OpenStack Volumes
```
Evan-Adam: retrieve OpenStack Volumes information instead ?

src/cloud/openstack/restapi/mode/volume.pm
scresto31: Yes, that's true, it's already done but I forgot to update the documentation

src/cloud/openstack/restapi/mode/volume.pm
scresto31: Yes, in the other modes I used "Manage", I will update it

src/cloud/openstack/restapi/mode/volume.pm
scresto31: In this case the users would have to use a dot but in fact "availability zones" (and other fields) are codes that shouldn’t contain a comma

src/cloud/openstack/restapi/custom/api.pm
```diff
@@ -857,6 +937,37 @@ Set the endpoint to use in the Glance service URL (default: '/v2').
 Allow insecure TLS connection (default: '0').
 When set to 0 the default insecure value passed with --insecure is used.
 
+=item B<--image-url>
```
Evan-Adam: forgot to change to volume ?
Also why do you duplicate these argument instead of using only one --url ?

tests/cloud/openstack/restapi/volume.robot
scresto31: Ok

src/cloud/openstack/restapi/custom/api.pm
scresto31: As seen in meet there can be different URLs depending on the services

PR #5868, opened 2025-11-28
CTOR-1916 Plugin(cloud::openstack::restapi) - Mode(network)

scresto31:
## Description

Plugin(cloud::openstack::restapi) - Mode(network)

**Fixes** # CTOR-1916

## Type of change

### DIFF
--- a/src/cloud/openstack/restapi/custom/api.pm
+++ b/src/cloud/openstack/restapi/custom/api.pm
@@ -93,7 +93,8 @@ sub new {
         _service_filters_options(type => 'volume', ),
         _service_ident_options(type => 'network',   port => 9696, endpoint => ''),
         _service_filters_options(type => 'network', ),
-
+        _service_ident_options(type => 'loadbalancer',    port => 9876, endpoint => ''),
+        _service_filters_options(type => 'loadbalancer', ),
 
         # Endpoint filters common to all services
         _endpoint_filters_options(),
@@ -148,6 +149,7 @@ sub service_check_options {
 
     $self->{$service.$_} = $self->{option_results}->{$service.$_}
         foreach qw/url insecure endpoint port interface timeout/;
+
     # Add default endpoint to URL if defined and not already present
     $self->{$service.'url'} .= $self->{$service.'endpoint'}
         if $self->{$service.'url'} ne '' && $self->{$service.'endpoint'} ne '' && $self->{$service.'url'} !~ /$self->{$service.'endpoint'}$/;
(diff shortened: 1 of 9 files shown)


src/cloud/openstack/restapi/custom/api.pm
```diff
@@ -446,6 +450,96 @@ sub nova_list_instances {
...
+    my @results;
+
+    # Retry to handle token expiration
+    while (1) {
+        $response_brut = $self->{http}->request(
+            method => 'GET',
+            get_params => \%params,
+            header => [ 'Content-Type: application/json',
+                        'X-Auth-Token: '.$token],
+            $self->connect_info( url => $self->{network_url},
```
Evan-Adam, line 663: Why don't you put the v2.0 part in the parameter l94 ? So when they do a v3 if it's compatible users can update without waiting for us.

src/cloud/openstack/restapi/custom/api.pm
```diff
@@ -89,8 +89,11 @@ sub new {
         _service_filters_options(type => 'compute', ),
         _service_ident_options(type => 'image',    port => 9292, endpoint => ''),
         _service_filters_options(type => 'image', ),
-        _service_ident_options(type => 'volume',   port => 9292, endpoint => ''),
+        _service_ident_options(type => 'volume',   port => 8776, endpoint => ''),
         _service_filters_options(type => 'volume', ),
+        _service_ident_options(type => 'network',   port => 9696, endpoint => ''),
```
Evan-Adam, line 94: I'm not sure about this way to manage the arguments given by the user, I would expect a single --hostname, --url, --insecure instead of duplicating it for every mode, but if it's the right way to do you need to document the new arguments in this help

src/cloud/openstack/restapi/mode/network.pm
```diff
@@ -0,0 +1,337 @@
+#
```
Evan-Adam, line 1: Did you add the test in another PR ?

src/cloud/openstack/restapi/custom/api.pm
Evan-Adam: Discuted in vocal : There will be one host openstack with at least one service per tool in openstack. As openstack can be distributed, we need to manage the case where some service have to access a different url.

Evan-Adam, line 94: Miss documentation of these options in pod

Evan-Adam (changes requested):
miss some documentation for the network mode (maybe you can make it generic as *-url seeing the size of the help message)

PR #5873, opened 2025-12-01
Update ntp.pm

cedr-x:
offset with a negative "-1 second" output cause a critical status of the plugin.  Offsets should be positive integers

# Community contributors

## Description

Sometimes, centreon_linux_snmp.pl --plugin=os::linux::snmp::plugin --mode=time trigger critical status when the checked host is just a second behind the ref_time
```
sudo -u centreon-engine /usr/lib/centreon/plugins//centreon_linux_snmp.pl --plugin=os::linux::snmp::plugin --mode=time --hostname=192.168.1.118 --snmp-version=2c --snmp-community=****** --warning-offset=120 --critical-offset=300 
CRITICAL: Time offset -1 second(s): Local Time : 2025-12-01T09:42:30 (Europe/Paris) | 'offset'=-1s;0:120;0:300;;
```

## Type of change

- Patch fixing an issue (non-breaking change)

Check against an host which have it clock set near 1s behing ntp ref time
The unpatched check is tossed be criticals when it retrives exactly "-1" (second) as offset
This patched check, returning in that case "1" stay with status OK

tested with:

```
sudo -u centreon-engine /usr/lib/centreon/plugins//centreon_linux_snmp.pl --plugin=os::linux::snmp::plugin --mode=time --hostname=192.168.1.118 --snmp-version=2c --snmp-community=****** --warning-offset=120 --critical-offset=300 
OK: Time offset 1 second(s): Local Time : 2025-12-01T09:51:12 (+0100) | 'offset'=1s;0:120;0:300;;
```

## Checklist

- [x] I have commented my code, especially **hard-to-understand areas** of the PR.
- [x] I have provide data or shown output displaying the result of this code in the plugin area concerned.

### DIFF
--- a/src/snmp_standard/mode/ntp.pm
+++ b/src/snmp_standard/mode/ntp.pm
@@ -173,7 +173,7 @@ sub manage_selection {
         $ref_time = time();
     }
 
-    my $offset = $distant_time - $ref_time;
+    my $offset = abs( $distant_time - $ref_time);
     my $remote_date_formated = sprintf(
         'Local Time : %02d-%02d-%02dT%02d:%02d:%02d (%s)',
         $remote_date->[0], $remote_date->[1], $remote_date->[2],


omercier:
Hi @cedr-x,
Thanks for your contribution, but I think losing the minus signs would mean that you don't know if your host is ahead or late compared to the reference time.

Here's what I suggest:
- **Use Nagios ranges**: setting the thresholds this way `--warning-offset=-120:120` will return a warning status only when the value is outside the interval [-120:120]. More information [here](https://nagios-plugins.org/doc/guidelines.html#thresholds). You can use this option without changing the code.
- **Smart guessing the user's intentions**: instead of transforming the threshold into its absolute value, suppose that if the threshold value is a positive integer, then the user probably meant "I want the offset's absolute value to be less than x", and silently change the threshold from `n` to `-n:n`.

What do you think?

omercier:
Hi @cedr-x,
As I said your pull request cannot be accepted as it is, so I'm closing it.
I have however opened another PR #6076 to update the help output and the official documentation.
Feel free to open a new one if you want to implement the second proposition I made:

> **Smart guessing the user's intentions**: instead of transforming the threshold into its absolute value, suppose that if the threshold value is a positive integer, then the user probably meant "I want the offset's absolute value to be less than x", and silently change the threshold from n to -n:n.

Thanks anyway for drawing our attention on this issue 💯 
Best regards

lucie-tirand:
Hello :)

Thanks for this contribution again ! It's handled in the last release 26.04 : [plugins-20260400](https://github.com/centreon/centreon-plugins/releases/tag/plugins-20260400).

PR #5897, opened 2025-12-12
CTOR-2036 Plugin(apps::backup::commvault::commserve::restapi) - Mode(jobs): add client name to output prefix

scresto31:
# Centreon team (internal PR)

## Description

Plugin(apps::backup::commvault::commserve::restapi) - Mode(jobs): add client name to output prefix

**Fixes** # CTOR-2036

## Type of change

### DIFF
--- a/src/apps/backup/commvault/commserve/restapi/mode/jobs.pm
+++ b/src/apps/backup/commvault/commserve/restapi/mode/jobs.pm
@@ -25,7 +25,7 @@ use base qw(centreon::plugins::templates::counter);
 use strict;
 use warnings;
 use Digest::MD5 qw(md5_hex);
-use centreon::plugins::misc;
+use centreon::plugins::misc qw/is_excluded/;
 use centreon::plugins::templates::catalog_functions qw(catalog_status_threshold_ng);
 
 sub custom_status_output {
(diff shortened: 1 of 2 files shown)


src/apps/backup/commvault/commserve/restapi/mode/jobs.pm
```diff
@@ -43,7 +43,11 @@ sub custom_long_output {
 sub prefix_job_output {
     my ($self, %options) = @_;
 
-    return "Job '" . $options{instance_value}->{display} . "' [type: " . $options{instance_value}->{type} . "] " ;
+
+    my $client_part = $options{instance_value}->{client_name};
```
Evan-Adam, line 47: If the $options{instance_value}->{client_name} value is "notAvailable" you will print it on the output, I think you should not add anything in this case

src/apps/backup/commvault/commserve/restapi/mode/jobs.pm
```diff
@@ -43,7 +43,11 @@ sub custom_long_output {
...
+    $client_part = "client: $client_part, " if $client_part ne '' && $client_part ne 'notAvailable';
+
+    return "Job '" . $options{instance_value}->{display} . "' [" . $client_part . "type: " . $options{instance_value}->{type} . "] " ;
```
Evan-Adam, line 50: you can add a hyphen or a underscore to separate both value if there is any $client_part to use
I think this hyphen should be in the $client_part to avoid adding it if the variable is empty

src/apps/backup/commvault/commserve/restapi/mode/jobs.pm
scresto09: Hum, normally the "notAvailable" is filtered out with test line 48

src/apps/backup/commvault/commserve/restapi/mode/jobs.pm
scresto09: When there is a client the output is "Job: XX [client: test, type: something]"
When there is no client the output is "Job: XX [type: something]"
So I think a hyphen is probably not necessary, no ?

PR #5906, opened 2025-12-17
Veeam Enterprise Manager REST API plugin not working for new version 13

rmorandell-pgum:
# Community contributors

## Description

I have the following version with the following license.

image

License

image

I'm not getting any results from the plugin because the body looks different. In the body I get back, everything is in lowercase. 
If I adjust the plugin accordingly, it works.

image

When I change in the --list-repositories mode all to lower case works.

image

**Fixes** # (issue)

I have modified the function so that after the JSON, the code recursively goes through the content and sets all keys to lowercase. Accordingly, the “manage_selection” functions in the modes have been adjusted and the lowercase keys are also used. This makes the plugin compatible with both old and new versions.

## Type of change

Can be tested with previous test data, as nothing changes in the logic, only the access to the keys in the refs.

### DIFF
--- a/src/apps/backup/veeam/vbem/restapi/custom/api.pm
+++ b/src/apps/backup/veeam/vbem/restapi/custom/api.pm
@@ -208,6 +208,7 @@ sub request_api {
     my $decoded;
     eval {
         $decoded = JSON::XS->new->decode($content);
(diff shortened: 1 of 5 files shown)


omercier:
Thanks @rmorandell-pgum 
Code pulled in PR #6005

PR #6127, opened 2026-04-15
Fix database mssql mode deadlocks

rmorandell-pgum:
# Community contributors

## Description

According to my analysis, the mode for database::mssql::mode::lockswaits and database::mssql::mode::deadlocks is incorrectly designed.

Here is the Microsoft documentation about this performance counters.
 
1.	Deadlocks cannot be filtered at the database level using performance counters. The instance here is not the DB instance or the database, but a lock type.

image
 
In my case the query of the plugin executed on the MSSQL Server looks like this. 

 
image

There is always a “_Total”. So, the filter should be –filter-instance instead of –filter-database and the default value should be “_Total” which is the total of all other single instances.

2. The counter “Lock Waits/sec” isn’t a real time counter. You would expect the value to represent “per second”, but instead the number keeps increasing continuously. Although the counter is named Lock Waits/sec, the value returned in sys.dm_os_performance_counters is a cumulative raw counter, not an already calculated per-second rate. cntr_value stores the total number of lock waits since SQL Server was last started. It continues to increase over time. It resets only when the SQL Server service restarts.

This gives us the difference from the last measurement.

image

## Type of change

- [x] Breaking change (patch or feature) that might cause side effects breaking part of the Software

## Checklist

- [ ] I have commented my code, especially **hard-to-understand areas** of the PR.
- [ ] I have **rebased** my development branch on the base branch (develop).
- [ ] I have provide data or shown output displaying the result of this code in the plugin area concerned.

## Summary by Aikido
|  Security Issues: 0 | 🔍 Quality Issues: 2 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**⚡ Enhancements**
* Replaced --filter-database with --filter-instance defaulting to '_Total'.
* Converted cumulative performance counters to per-second and added totals.
* Enabled statefile caching and generated unique cache names using md5.
* Added verbose debug messages when skipping or reporting performance instances.

<sup>[More info](https://app.aikido.dev/featurebranch/scan/104988467?groupId=59752)</sup>


omercier:
Hi @rmorandell-pgum,
Do you agree what is done [here](https://github.com/centreon/centreon-plugins/pull/6230)?

lucie-tirand:
Hello :)

Thanks for your contribution 🙏
This issue is handled in the [latest release 20260600](https://github.com/centreon/centreon-plugins/releases/tag/plugins-20260600).

Kind regards.

PR #6246, opened 2026-06-17
fix(hooks): only check robocop when .robot files are committed

pkippes:
## Summary

The `robocop` check was running unconditionally at the top of the script, causing a fatal error even when no `.robot` files are being committed (e.g. a simple version bump).

Moved the `robocop` detection inside the `robot)` case block so it only runs when actually needed.

## Root cause

Introduced in commit `b9fef835d` — the upfront check was added alongside the `robot)` case but was placed at the top level.


scresto31:
dpulicate => https://github.com/centreon/centreon-plugins/pull/6247

pkippes:
duplcated

PR #6258, opened 2026-06-19
enh(network::stormshield::snmp): added and updated multiple modes

crsuser:
# Community contributors

## Description

This pull request introduces three new operational modes and significantly enhances several existing ones.

The new modes added are:
- `auto_update`: Handles automatic update checks and procedures.
- `ha_cluster`: Provides monitoring capabilities for High Availability clusters.
- `interfaces_disco`: A dedicated **discovery mode** designed to automatically detect and enumerate network interfaces on the target device. This mode simplifies the configuration process by dynamically identifying available interfaces rather than requiring manual definition.
- `router_disco`: A dedicated **discovery mode** designed to automatically detect and aggregate gateways for Stormshield routers. This mode simplifies monitoring by dynamically identifying principal and backup paths, analyzing their latency and packet loss, and reporting a unified health status for the entire routing configuration.

Existing modes have also been updated to integrate new MIBs and data points, ensuring broader and more granular supervision coverage.

## Type of change

To verify the new functionality, please follow these steps:

1. **Test `interfaces_disco` (Discovery Mode):**
   - Run the plugin with the `--mode=interfaces-disco` option against a target device.
   - Verify that the output lists all detected network interfaces dynamically.
   - In Centreon UI, use the service scan option.

2. **Test `router_disco` (Discovery Mode):**
   - Run the plugin with the `--mode=router-disco` option against a target device.
   - Verify that the output lists all detected routers and gateways dynamically.
   - In Centreon UI, use the service scan option.

3. **Test `auto_update`:**
   - Execute the plugin with `--mode=auto-update`.
   - Verify the update status report.

4. **Test `ha_cluster`:**
   - Run `--mode=ha-cluster` against a configured HA cluster.
   - Validate that cluster node states and synchronization status are reported correctly.

## Checklist

- [x] I have commented my code, especially **hard-to-understand areas** of the PR.
- [x] I have provided data or shown output displaying the result of this code in the plugin area concerned.

### DIFF
--- a/src/network/stormshield/snmp/mode/memorydetailed.pm
+++ b/src/network/stormshield/snmp/mode/memorydetailed.pm
@@ -1,5 +1,5 @@
 #
-# Copyright 2024 Centreon (http://www.centreon.com/)
+# Copyright 2026-Present Centreon (http://www.centreon.com/)
 #
 # Centreon is a full-fledged industry-strength solution that meets
 # the needs in IT infrastructure and application monitoring for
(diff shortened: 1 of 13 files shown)


src/network/stormshield/snmp/mode/memorydetailed.pm
```diff
@@ -24,151 +24,284 @@ use base qw(centreon::plugins::templates::counter);
...
 
-sub prefix_memory_output {
-    my ($self, %options) = @_;
-
-    return 'Memory usage ';
-}
+my @mem_labels = qw(asq icmp frag host system dtrack socket etherstate user);
 
 sub set_counters {
     my ($self, %options) = @_;
```
Evan-Adam, line 33: Hello @crsuser , thanks for the contribution !
I'm in the process of integrating changes in this file to the perl plugins and our monitoring connector.
I see you removed the prefix_memory_output() function and all the nlabel from the counter, could you expand a bit on why you wanted to remove them ?

src/network/stormshield/snmp/mode/memorydetailed.pm
```diff
@@ -24,151 +24,284 @@ use base qw(centreon::plugins::templates::counter);
...
+                        value => 'frag',
+                        template => '%s',
+                        unit => '%',
+                        min => 0,
+                        max => 100,
+                    }
                 ]
             }
         },
-        { label => 'conn', nlabel => 'memory.connections.percentage', set => {
```
Evan-Adam, line 66: you removed the conn metric from the plugin, but for user using legacy stormshield and new centreon version this will remove a data point they already have.
I see in l228 the oid snsMemConn that you translate later to "asq", is it the same concept ? if yes we can use the same name "conn" to avoid a breaking change.

src/network/stormshield/snmp/mode/memorydetailed.pm
```diff
@@ -24,151 +24,284 @@ use base qw(centreon::plugins::templates::counter);
...
-                key_values => [ { name => 'conn' } ],
-                output_template => 'connections: %.2f %%',
+        { label => 'host', set => {
+                key_values => [ { name => 'host' } ],
+                output_template => "Host: %s%%",
                 perfdatas => [
-                    { template => '%.2f', min => 0, max => 100, unit => '%' }
+                    {
+                        label => 'mem_host',
```
Evan-Adam, line 90: Changing our metric names is sadly considered a breaking change, so we can't add this "label" key.
If you want better perfdata name you can use `--use-new-perfdata` option which will enable the "nlabel", providing better metrics name

src/network/stormshield/snmp/mode/memorydetailed.pm
```diff
@@ -24,151 +24,284 @@ use base qw(centreon::plugins::templates::counter);
...
-    
+
     $self->{maps_counters_type} = [
-        { name => 'global', type => 0, cb_prefix_output => 'prefix_memory_output', skipped_code => { -10 => 1 } }
+        { name => 'global', type => COUNTER_TYPE_GLOBAL, skipped_code => { NO_VALUE() => 1 } }
     ];
-    
+
     $self->{maps_counters}->{global} = [
-        { label => 'total', nlabel => 'memory.usage.percentage', set => {
```
Evan-Adam, line 42: why did you remove the total metric ? was the computation wrong ?

src/network/stormshield/snmp/mode/memorydetailed.pm
```diff
@@ -24,151 +24,284 @@ use base qw(centreon::plugins::templates::counter);
...
+            $mem_values{'host'}       = $values[0];
+            $mem_values{'frag'}       = $values[1];
+            $mem_values{'icmp'}       = $values[2];
+            $mem_values{'asq'}        = $values[3];
+            $mem_values{'etherstate'} = $values[4];
+            $mem_values{'dtrack'}     = $values[5];
+            $mem_values{'system'}     = $values[6];
+
+            if ($fields == 8) {
+                $mem_values{'socket'} = $values[7];
```
Evan-Adam, line 265: does the meaning of the legacy oid have changed ?
I see in the old code we expecte either 6 or 7 field, with the optionnal one on the middle, your change expect 7 or 8 of them ?
does this behaviour changed in a previous stormshield version or was this always the case ?

src/network/stormshield/snmp/mode/memorydetailed.pm
crsuser: The total metric was removed because it isn't relevant in this context. Each metric (memory percentage) is independent and can range from 0% to 100% on its own. Summing them would result in a total that could exceed 100%, which would be misleading and not provide meaningful insight.

src/network/stormshield/snmp/mode/memorydetailed.pm
crsuser: Hello! Thanks for the quick response.

The changes to the plugin were prompted by a new version of Stormshield firewalls, which introduced new MIBs and OIDs. Specifically, all versions post 4.8.9 now fetch metrics via a table, rather than using a string containing all metrics separated by ", " as in previous versions.

Given this architectural shift, I took the opportunity to refactor the file. As part of this overhaul, I removed the prefix_memory_output() function as I considered it less relevant in the output.

src/network/stormshield/snmp/mode/memorydetailed.pm
crsuser: Yes, snsMemConn corresponds to the OID that we renamed ASQ in the plugin. To be more specific, it is a combination of the two : it refers to the percentage of memory usage on ASQ connections.
I renamed it to align with current SNS terminology. If you consider this as a breaking change, you can keep the conn name. It's totally fine for us !

However, when testing the commits you made today, the ASQ/Conn metric is no longer being returned.
This is because, in the perfdatas, the key_values name remained “asq” instead of “conn”.
I've submitted a review to fix this in omercier's PR.

src/network/stormshield/snmp/mode/memorydetailed.pm
crsuser: Understood on the constraint. This was added to give clearer names, without realizing this counts as a breaking rename rather than an addition.
I saw in your commits that you're using nlabels, that works perfectly for me.

src/network/stormshield/snmp/mode/memorydetailed.pm
crsuser: The 6-7 field expectation in the old file was always wrong. It was a longstanding bug there, not a behavior change on the Stormshield side.
The correct field is 7 or 8.

lucie-tirand:
Hello :)

Thanks for your contribution 🙏 
This PR is handled by the [26.08 release](https://github.com/centreon/centreon-plugins/releases/tag/plugins-20260800).

PR #6264, opened 2026-06-23
enh(rust-plugins) Update package name + destination

scresto31:
## Description

Update package name + destination

**Fixes** # CTOR-2281

## Type of change

- [x] Breaking change (patch or feature) that might cause side effects breaking part of the Software


.github/workflows/generic-plugins.yml
```diff
@@ -181,7 +181,7 @@ jobs:
               grep " 0100007F:07E8 " /proc/net/udp >/dev/null && break
               sleep 0.5
           done ; echo "snmpsim is listening"' || exit 1
-          robot --include centreon-generic-snmp --exclude notauto --outputdir /tmp/robot-tests ./tests/os/linux/snmp/
+          robot --include centreon-plugin-rust-snmp --exclude notauto --outputdir /tmp/robot-tests ./tests/os/linux/snmp/
```
tuntoja, line 184: shouldn t there also be a test on rs-collections/operatingsystems-linux-snmp content, since its was also added to packaging, it means it will make it to stable at some point ?

.github/workflows/plugins.yml
tuntoja, line 47: should we add the new rs-collection as a trigger (else workflow will not trigger if there is only changes on rs-collections)

.github/workflows/generic-plugins.yml
omercier: It should not have been added yet, I'll remove it to introduce it a bit later

.github/workflows/plugins.yml
omercier: Yes we should

omercier (approved):

PR #6384, opened 2026-08-31
fix(docker-connector-vmware): dockerfile-review fixes

Tpo76:
## Summary

`/dockerfile-review` audit of `.github/docker/connector/Dockerfile.connector-vmware`, plus follow-up fixes from review discussion. Every change below was verified with a real local `docker build`/`docker run` (and, where noted, against the real licensed VMware SDK), not just read statically.

### Security / correctness fixes
- **`useradd` missing `-l`** for `centreon`/`centreon-engine`/`centreon-gorgone` — added.
- **Pipes without `pipefail`**: two `wget | gpg --dearmor | tee` apt-key fetches (in `extractor-repo` and the runtime stage) ran under `bash -e`, which only checks the last command's exit code — a failed `wget` would silently produce a broken/empty keyring instead of failing the build. Switched to `bash -eo pipefail`.
- **`gnupg`/`wget` never removed**: installed to bootstrap the Centreon apt repo, but shipped permanently in the production image. Now purged in the same `RUN` (confirmed via `dpkg -l` they're gone, and nothing in the entrypoint scripts needs them).
- **Merged `extractor-repo`'s two separate `apt-get update && install` blocks** into one.

### Secret handling
- Removed a redundant `ENV APP_SECRET=/SALT=/DEBUG=` documentation block from the Dockerfile — the entrypoint's own `:?` requirement checks already document this, and the empty-valued ENV was tripping Docker's own `SecretsUsedInArgOrEnv` build warning.
- Found and fixed a real leak while testing that removal: `container.d/10-engine-context.sh` is *sourced* (not executed) by `container.sh`, so `DEBUG=true`'s `set -x` was inherited into it and printed `APP_SECRET`/`SALT`'s actual values in the trace (`+ APP_SECRET=...`, `+ printf ... <secret>`) — on top of an explicit `cat` of the secrets file it also did. Both fixed: tracing is disabled around the secret-handling block (keyed directly off the same `DEBUG` check `container.sh` uses) and restored afterward so later scripts still trace normally; the debug line now shows file permissions instead of content.

### VERSION handling
- `ARG VERSION`/`ARG STABILITY` fed `LABEL org.opencontainers.image.version`, but the CI workflow's `build-args` never passed them — the label was always empty. Wired `VERSION`/`STABILITY` from `get-environment` outputs into `connector-vmware.yml`'s `docker-build` job.
- The README already documented `--build-arg VERSION=...` as pinning a specific release, but the Dockerfile never actually used it for anything except the label — `apt-get download` always grabbed latest-in-channel regardless. Now `apt-get download "pkg${PKG_VERSION:+=$PKG_VERSION}"` actually pins when set, falls back to latest when empty.
- Caught a real bug while wiring that up: naming the shell variable `VERSION` broke the *unpinned* build (`E: Version '13 (trixie)' ... was not found`) because `. /etc/os-release` in the same block defines its own `VERSION="13 (trixie)"`, silently clobbering the ARG-sourced one. Renamed to `PKG_VERSION`.

### Verification stage
- Added a `WITH_SDK`-conditional functional check instead of just checking file existence: `perl -c /usr/bin/centreon_vmware.pl` when `WITH_SDK=true` (real compile/dependency check — catches a broken `PERL5LIB`, a bad SDK patch, a missing runtime Perl dep); an assertion that the daemon's own graceful "you will need the Perl VMware SDK" message still fires when `WITH_SDK=false`, instead of silently accepting any other failure mode.
- This surfaced a stale/incorrect comment: the `sdk-false` stage claimed "the daemon starts but encrypted credentials won't be decryptable without the SDK." Empirically false — `centreon::script::centreon_vmware` unconditionally `use`s `VMware::VIRuntime`/`VMware::VILib`, so the daemon can't even compile without the SDK, plain-text or encrypted. Comment corrected, and the same wrong claim in `sdks-vmware/README.md`'s "Local build without SDK" section fixed too.

### Local dev defaults
- `connector-vmware.yml` always passes `PACKAGE_SOURCE`/`WITH_SDK` explicitly, so the Dockerfile's own `ARG` defaults only ever affected a plain `docker build` with no flags. The old defaults (`mount` + `false`) produced either a build error (`packages-centreon/` doesn't exist on a fresh clone) or a technically-valid but functionally useless image (daemon can't run at all without the SDK). Flipped defaults to `PACKAGE_SOURCE=repo`, `WITH_SDK=true` (all declaration sites) so a plain `docker build -f Dockerfile.connector-vmware .` either builds a fully working image or fails immediately pointing at the missing SDK files.
- Updated `sdks-vmware/README.md` to match: corrected a stale reference to a `docker-builder-connector-vmware.yml` workflow that doesn't exist, dropped now-redundant `--build-arg` flags from the example commands, and documented `--build-arg VERSION=$(cat .version.plugins)` for `local`/`mount` builds (which don't download a `.deb`, so there's no packaged version to read from).

### Not changed (reported, not fixed)
- **Unpinned package versions**: come from live Debian/Centreon apt repos; pinning needs a resolved-version snapshot first, not a guess.
- **Apt source-list `echo > file` lines**: left as `RUN echo` rather than a COPY heredoc — content depends on `VERSION_CODENAME` sourced from `/etc/os-release` at build time, so a static heredoc isn't a direct swap.
- **docker-compose `build:` support** for `centreon-vmware` in `centreon/centreon`'s `install-poller` (different repo) — tracked separately in [MON-208038](https://centreon.atlassian.net/browse/MON-208038).

## Test plan

- [x] Built locally end-to-end with `PACKAGE_SOURCE=local`/`mount`/`repo` and both `WITH_SDK=true` (real licensed SDK) and `WITH_SDK=false`
- [x] Confirmed via `dpkg -l` that `gnupg`/`wget` are absent from the final image
- [x] Confirmed `useradd -l` took effect (`/etc/passwd` entries correct, UIDs/GIDs unchanged: 900/901/903)
- [x] Confirmed `DEBUG=true` no longer prints `APP_SECRET`/`SALT` anywhere in the trace; normal and missing-secret paths unchanged
- [x] Confirmed `VERSION` pinning: real version → build succeeds and label matches exactly; bogus version → build fails clearly; unpinned → downloads latest
- [x] Confirmed the verification check passes on `WITH_SDK=true` (`perl -c` "syntax OK") and `WITH_SDK=false` (graceful-failure assertion), and genuinely fails when the expected message is missing (isolated repro)
- [x] Confirmed a plain `docker build` with zero flags now either builds a fully working image or fails immediately pointing at the missing SDK
- [x] Tested on both amd64 and arm64 (QEMU emulation) with the real SDK
- [x] Real CI (`connector-vmware.yml`, `docker build connector-vmware trixie`) passed against the final state of this branch

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[MON-208038]: https://centreon.atlassian.net/browse/MON-208038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/docker/connector/Dockerfile.connector-vmware
+++ b/.github/docker/connector/Dockerfile.connector-vmware
@@ -45,7 +45,7 @@ ARG STABILITY=stable
 
 RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     --mount=type=cache,target=/var/lib/apt/lists,sharing=locked \
-    STABILITY=${STABILITY} bash -eo pipefail <<'EOF'
+    STABILITY=${STABILITY} PKG_VERSION=${VERSION} bash -eo pipefail <<'EOF'
 apt-get update
 apt-get install -y --no-install-recommends dpkg gnupg wget ca-certificates
 . /etc/os-release
@@ -55,7 +55,11 @@ wget -O- https://apt-key.centreon.com | gpg --dearmor \
 apt-get update
 mkdir -p /extracted
 cd /tmp
-apt-get download centreon-plugin-virtualization-vmware-daemon
+# PKG_VERSION pins an exact release (e.g. 20260300-1+deb13u1); unset/empty
+# downloads whatever is currently latest in the ${STABILITY} channel.
+# (named PKG_VERSION, not VERSION: `. /etc/os-release` above already defines
+# its own VERSION="13 (trixie)", which would silently clobber an ARG VERSION.)
+apt-get download "centreon-plugin-virtualization-vmware-daemon${PKG_VERSION:+=$PKG_VERSION}"
 for deb in centreon-plugin-virtualization-vmware-daemon*.deb; do
     echo "Extracting $deb..."
     dpkg-deb -x "$deb" /extracted/
--- a/.github/docker/connector/Dockerfile.connector-vmware
+++ b/.github/docker/connector/Dockerfile.connector-vmware
@@ -269,12 +269,6 @@ EOF
 
 COPY --chmod=755 .github/docker/connector/entrypoint /usr/local/lib/centreon-vmware/
 
-# Required at runtime by entrypoint/container.d/10-engine-context.sh (poller mode).
-# DEBUG (true/1) enables verbose entrypoint logging — see container.sh / 00-init.sh.
-ENV APP_SECRET= \
-    SALT= \
-    DEBUG=
-
 ENV PERL5LIB=/usr/local/share/perl5:/usr/share/perl5:/usr/lib/perl5
 
 EXPOSE 5700
--- a/.github/docker/connector/Dockerfile.connector-vmware
+++ b/.github/docker/connector/Dockerfile.connector-vmware
@@ -1,11 +1,14 @@
 # syntax=docker/dockerfile:1
 
-# PACKAGE_SOURCE=mount (default, CI): extract from .deb bind-mounted via packages-centreon/
-# PACKAGE_SOURCE=repo  (prod):        download from Centreon apt repo (requires VERSION + STABILITY)
-# PACKAGE_SOURCE=local (dev):         copy from local repo source (connectors/vmware/src/)
-ARG PACKAGE_SOURCE=mount
-# Build arg to include VMware Perl SDK (requires proprietary files in sdks-vmware/ — see sdks-vmware/README.md)
-ARG WITH_SDK=false
+# PACKAGE_SOURCE=repo  (default, local dev): download from Centreon apt repo (VERSION pins a release, STABILITY picks the channel)
+# PACKAGE_SOURCE=mount (CI):                 extract from .deb bind-mounted via packages-centreon/
+# PACKAGE_SOURCE=local (Dockerfile-only CI): copy from local repo source (connectors/vmware/src/)
+ARG PACKAGE_SOURCE=repo
+# Build arg to include VMware Perl SDK (requires proprietary files in sdks-vmware/ — see sdks-vmware/README.md).
+# Defaults to true: without the SDK the daemon can't do anything useful (see the
+# sdk-false stage below) — CI always passes WITH_SDK=false explicitly to validate
+# the Dockerfile builds without the licensed SDK, it doesn't rely on this default.
+ARG WITH_SDK=true
 
 #############################################
 # Stage 1a: Extract from bind-mounted .deb (PACKAGE_SOURCE=mount)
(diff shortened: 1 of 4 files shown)


Tpo76:
Addressed review feedback:

- Removed the `ENV APP_SECRET=/SALT=/DEBUG=` block from the Dockerfile — redundant with the entrypoint's own `:?` requirement checks, and it triggered Docker's own `SecretsUsedInArgOrEnv` build warning.
- Found and fixed a deeper issue while testing: `container.d/10-engine-context.sh` is *sourced* by `container.sh`, so `DEBUG=true`'s `set -x` was inherited and printed the actual `APP_SECRET`/`SALT` values in the trace (`+ APP_SECRET=...`, `+ printf ... <secret>`), independently of the explicit `cat` of the secrets file it also did. Both are now fixed: tracing is disabled around the secret-handling block and restored afterwards, and the debug output shows file permissions instead of file content.

Verified locally: `DEBUG=true` no longer prints secret values anywhere in the trace, the container still reaches "ready", and the missing-`APP_SECRET`/`SALT` failure path is unchanged.

Tpo76:
Added a real functional check to the verification stage, per a question about whether \`perl -c\` was worth adding.

Turns out \`centreon::script::centreon_vmware\` unconditionally \`use\`s \`VMware::VIRuntime\`/\`VMware::VILib\` — so a plain \`perl -c\` would **always** fail on the CI-default \`WITH_SDK=false\` build (verified: the actual daemon CMD exits 29 immediately without the SDK, contradicting the old \`sdk-false\` stage comment claiming "the daemon starts"). Made the check conditional instead:

- \`WITH_SDK=true\`: \`perl -c /usr/bin/centreon_vmware.pl\` must succeed (real regression check — catches a broken \`PERL5LIB\`, a bad SDK patch, a missing runtime Perl dep).
- \`WITH_SDK=false\`: asserts the daemon's own BEGIN-time requirements check still fires its documented \`"you will need the Perl VMware SDK"\` message, instead of silently accepting any other failure mode.

Also fixed the stale \`sdk-false\` comment and moved \`ENV PERL5LIB\` earlier so \`perl\` can actually resolve the copied modules during verification.

Verified locally: \`WITH_SDK=false\` build passes (graceful-failure assertion holds), \`WITH_SDK=true\` build passes (\`syntax OK\`), and an isolated repro confirms the check genuinely fails when the expected message is missing (not a check that passes unconditionally).

kduret (approved):

PR #6421, opened 2026-09-04
feat(rust-plugins): SNMPv3 with the User-based Security Model

julienmathis:
## ⚠️ Not yet interop-tested

Unlike the rest of this series, this PR has **not** been validated against a real SNMPv3 agent — no v3 device/simulator was available in the environment this was written in. The RFC 3414 §A.3 test vectors confirm the key-derivation math is byte-correct, and the DES/AES roundtrip tests confirm internal self-consistency, but **neither proves wire compatibility** with a real agent (e.g. against net-snmp as an independent oracle). That requires a dedicated SNMPv3 test setup and should happen before merging this one — please treat this as a draft for review/testing help rather than merge-ready.

## Summary
`rasn-snmp` provides the v3 ASN.1 structures; everything that makes them *secure* is implemented in the new `src/snmp/usm.rs`:

| Area | Detail |
|---|---|
| Key derivation | password-to-key over the 1 MB expansion (RFC 3414 A.2) + localization to the authoritative engine |
| Authentication | truncated HMAC over the whole message — **MD5, SHA-1** (RFC 3414) and **SHA-224/256/384/512** (RFC 7860) |
| Privacy | **CBC-DES** (RFC 3414 §8) and **CFB128-AES-128** (RFC 3826), unique salt per message |
| Session | engine discovery + time synchronization (RFC 3414 §4), done once per collection |
| Context | `--contextname` / `--contextengineid` (Perl parity) |

CLI mirrors the Perl options: `--snmp-username`, `--authprotocol`, `--authpassphrase`, `--privprotocol`, `--privpassphrase`. Misconfigurations fail fast: privacy without authentication, credentials without `--snmp-version=3`, half-specified protocol/passphrase pairs.

Note the RFC 7860 tag length for SHA-224 is **16 bytes, not 12**: the protocol names encode the truncation (`usmHMAC128SHA224` = 128 bits) — implemented as such here, worth double-checking in interop testing since it's an easy spot to get wrong.

An agent refusing credentials answers with an **unauthenticated** Report (it cannot sign with a key it does not share). That report is read *before* the digest check and its `usmStats` counter translated into a human cause — instead of a cryptic "malformed digest".

## Deliberate scope limit
**AES-192/256 are rejected with an explicit message** rather than implemented: they exist only as expired drafts with two incompatible key-extension variants (Blumenthal, Reeder); picking the wrong one fails to interoperate *silently*. Better a clear error than a mystery.

## Test plan
- [x] `cargo build --release` / `cargo test`: 99 passed, 0 failed (11 new), including the **RFC 3414 §A.3 key-localization vectors** for MD5 and SHA-1 pinned as unit tests (byte-exact match against the RFC's own worked example), plus DES/AES roundtrips and salt uniqueness
- [x] Manual: `--help` lists the new flags; `--snmp-version=3` without `--snmp-username`, half-specified auth/priv pairs, priv-without-auth, and `--privprotocol=AES256` all fail fast with a clear `UNKNOWN` message and exit 3; a discovery attempt against an unroutable address times out cleanly through the existing retry/collect-timeout machinery (no panic, no hang)
- [ ] **Interop against a real SNMPv3 agent** — not done yet, see caveat above


julienmathis:
The \`plugins\` / \`check-status\` checks are failing on \`test-plugins jammy amd64\` and \`test-plugins noble amd64\`, both with \`RuntimeError: Connector is not running.\` from \`tests/resources/connector.py\` (the Perl connector daemon binary is absent in these two runner environments, so every robot test that goes through \`ctn_send_to_connector\` fails outright).

This is unrelated to this PR's changes (SNMPv3/USM support, plus a GetRequest-vs-GetBulk protocol fix inherited from the rebase) — those only touch \`rust-plugins/src/*.rs\`, never the Perl connector or the robot test harness. It's also not new: this same \`test-plugins\` job has failed on every CI run of this branch since it was created, each time on a different distro/arch (\`bullseye arm64\` on 2026-09-04 and 2026-09-08, now \`jammy\`/\`noble amd64\`), always with the same connector-not-found root cause. Confirmed reproducible after two reruns.

Flagging as a known pre-existing CI flake in the connector test harness rather than something this PR introduced.

tuntoja:
please update with the base, a key delivery fix has been merged in develop to ensure cpan libs for tests do exist in unstable on packages.centreon.com
https://github.com/centreon/centreon-plugins/commit/6ca734cc3c217b7a81a08812225d69b2e45bf418

PR #6425, opened 2026-09-08
fix(vmware-vsphere8): restore performance counters on vSphere 9.1 (vStats API removed)

julienmathis:
# Centreon team

> **Note for the reviewer:** this branch has no JIRA ID yet — please tell me the ticket to use and I will rename the branch accordingly.

## Description

**All performance counters of the vSphere 8 connector return nothing on vCenter 9.1.** Inventory modes (status, health, count, list) keep working, which is exactly what made this hard to diagnose in the field: the plugin does not crash, it quietly reports nothing.

### Root cause

The connector reads every metric through the **vStats API** (`com.vmware.vstats.*`) — `/api/stats/acq-specs` to create and extend acquisition specifications, `/api/stats/data/dp` to read data points. That API was shipped as **Tech Preview** in vSphere 8.0 and has been **removed in vSphere 9.1**.

Broadcom's own words in the [VCF 9.1 Product Support Notes](https://techdocs.broadcom.com/us/en/vmware-cis/vcf/vcf-9-0-and-later/9-1/release-notes/vmware-cloud-foundation-9-1-0-0-release-notes/vcf-91-product-support-notes.html):

> the vStats APIs (previously in Tech Preview) are removed.

This was confirmed by Broadcom support, who reproduced it on a test vCenter 9.1.0.0: the backend service is simply **no longer registered**, so `GET` and `POST` on `acq-specs` fail identically, before the request body is even parsed.

### ⚠️ Two VMware-side problems worth flagging

**1. The 9.1 Swagger / API Explorer is wrong.** It still lists `acq-specs` and `acq-specs/{id}` in its tree although the backend has been removed. Anyone trusting the published documentation without actually executing the call will conclude the API is still there. This cost us real diagnosis time and will cost it to others — do not trust the 9.1 API Explorer on this namespace.

**2. The official replacement is not a drop-in, and not even in vCenter.** The [Real-Time Metrics API](https://developer.broadcom.com/xapis/realtime-metrics-api/latest/) (Prometheus-compatible, PromQL, up to 2-second granularity) ships with **VCF Operations**, not with vCenter. It requires the Real-Time Data Service component to be explicitly deployed, a VCF or VVF environment, and **JWT Bearer authentication** instead of a vCenter session — a rewrite of the integration, not a URL change. It is therefore **not an option for standalone vCenter customers**, and is out of scope here: it belongs to the TAP discussion.

We also carry our own share of the blame: vStats was documented by VMware as Tech Preview with an explicit *"VMware does not guarantee backwards compatibility and recommends against using them in production environments"* ([vStats REST API reference](https://developer.vmware.com/apis/vsphere-automation/latest/vstats/vstats/)). Building the whole metric collection on it was a known risk.

### Why the failure was silent

Three separate behaviours combined, in [`src/apps/vmware/vsphere8/custom/api.pm`](https://github.com/centreon/centreon-plugins/blob/develop/src/apps/vmware/vsphere8/custom/api.pm):

1. `get_all_acq_specs()` did `push @{...}, grep {...} @{$response->{acq_specs}}`. With no `acq_specs` key the dereference yields nothing, and the trailing guard forced an empty array — **no error**.
2. That empty array was then **written to the cache statefile** (`vsphere8_api_acq_specs_<rsrc_id>_<hash>`) *before* being validated, so the emptiness was replayed on every subsequent run, even against a healthy vCenter.
3. `get_stats()` found no `data_points`, emitted `add_option_msg("no data ... at the moment")` and returned `undef` — an **informational message, not a failure**.

And when the 404 body was empty or HTML instead of JSON, the plugin died on a raw `Can't use string ("") as a HASH ref` rather than a readable error.

## Type of change

### 1. Detect the removal instead of collecting nothing

`vstats_response_is_usable()` validates that the response actually carries an `acq_specs` **collection**, and the emptiness is no longer cached.

The check keys on the **presence of the collection, not on it being non-empty** — a healthy vCenter that has no acquisition specification yet returns `{"acq_specs": []}` and must keep working. This is what makes the fix safe.

### 2. Read the counters from the vim25 `PerformanceManager`

Broadcom confirmed that the historical [`PerformanceManager`](https://developer.broadcom.com/xapis/vsphere-web-services-api/latest/vim.PerformanceManager.html) of the [vSphere Web Services (VIM) API](https://techdocs.broadcom.com/us/en/vmware-cis/vcf/vcf-9-0-and-later/9-0/administration-sdks-cli-and-tools/about-vmware-cloud-foundation-development/core-vsphere-apis/web-services-vim-api.html) **still answers on 9.1**. The new `apps::vmware::vsphere8::custom::vim25` package speaks vim25 directly over `centreon::plugins::http`:

```
RetrieveServiceContent -> Login -> QueryPerfCounter -> QueryPerf -> Logout
```

It requires **neither the unmaintained VMware Perl SDK nor the `centreon_vmware` daemon** — the SOAP envelopes are built in the module.

New option, documented in the plugin help:

```
--metrics-source=auto|vstats|vim25     (default: auto)
```

With `auto`, vStats is used and vim25 takes over when the vCenter no longer serves it, so **the same command line works from 8.0 to 9.1**. `vstats` and `vim25` pin one source, for instance to make an unsupported version fail loudly.

### Three design choices that deserve the reviewer's attention

**No mode is modified.** Counter ids keep their vStats naming (`cpu.capacity.usage.HOST`). The resolution happens inside the vim25 module, so the 20+ mode files are untouched and the two sources are interchangeable.

**Units are never assumed.** They are read from `unitInfo.key` in the catalogue **the vCenter itself advertises**, at runtime, and converted to the unit each mode expects. The `%target_units` table was derived from the actual arithmetic of each mode — for example `esx/mode/memory.pm` multiplies `mem.capacity.usable.HOST` by `1024*1024` to obtain bytes, so it expects MB. A conversion **between incompatible dimensions is refused** rather than reporting a wrong value. The classic vim25 trap where the `percent` unit is expressed in *hundredths* of a percent is handled explicitly and tested.

**Counter naming has a documented fallback.** A counter is first looked up under its vStats name (`cpu.capacity.usage.<rollup>`). If the vCenter does not advertise it, a documented alias table is tried (`cpu.usagemhz`, `mem.consumed`, `disk.maxTotalLatency`, `power.power`, …) and **the substitution is reported in the long output** so it can be audited. If neither matches, the plugin says exactly what it looked for instead of returning nothing.

### 3. Security hardening

vim25 responses are untrusted input. `parse_xml()` disables entity expansion and external DTD/network fetching (`no_network`, `expand_entities => 0`, `load_ext_dtd => 0`), ruling out XXE and entity-expansion attacks. **Both are covered by regression tests** (a file-read canary and a billion-laughs payload).

The vim25 session is also closed on destruction: a session lives 30 minutes by default and each run opens one, so leaving them behind would pile sessions up against the vCenter cap (3000 by default, with a dedicated alarm since 9.1).

### 4. Packaging

`XML::LibXML` / `libxml-libxml-perl` added to the **four** vsphere8 packages. They all ship `apps/vmware/vsphere8/custom/` and only declared `perl(JSON)`, so the package would have broken at runtime. `perl(Date::Parse)` on the Vcsa package is preserved.

### Automated tests

```bash
perl -Isrc tests/apps/vmware/vsphere8/api.t
```

**38 assertions, up from 11.** They cover:

| Area | What is asserted |
|---|---|
| vStats detection | `{"acq_specs": []}` is **usable** (healthy 8.0–9.0 with no spec yet); `{}`, an error body, a non-hash body, `undef` and a non-list `acq_specs` are all detected as unusable |
| Error message | states the 9.1 removal, names the vCenter, reports the HTTP code, points to `--metrics-source=vim25` |
| Unit conversion | kiloBytes→MB, megaHertz→kHz, hundredths of a percent→percent, identity; cross-dimension and unknown units are **refused** |
| Counter resolution | direct vStats name, alias fallback, explicit error when unmappable; `host-35`→`HostSystem`, `vm-42`→`VirtualMachine`, unparsable id refused |
| XML hardening | external entity **not** resolved, nested entities **not** expanded, malformed and empty payloads return `undef` |

### Manual end-to-end verification

I do not have a 9.1 to hand, so I verified against two mock vCenters (sources at the end of this description). Results:

| Scenario | Result |
|---|---|
| 9.1, catalogue exposes `cpu.capacity.*` (kiloHertz) | `OK: CPU average usage is 40.00 %, used frequency is 3200000 kHz` |
| 9.1, catalogue exposes only `cpu.usagemhz` (megaHertz) | **identical values**, through the alias table + MHz→kHz conversion |
| 9.1, `memory` mode, kiloBytes vs megaBytes catalogues | `50% ... 64.00 GB used - Usable: 128.00 GB` in both cases |
| 9.1, `power` mode | `OK: Power usage is 210 Watts` |
| 9.1, `--metrics-source=vstats` | explicit `UNKNOWN`, no silent fallback |
| 9.1, `--metrics-source=bogus` | `UNKNOWN: Unsupported --metrics-source 'bogus'. Expected one of: auto, vstats, vim25.` |
| **9.0 healthy, default `auto`** | **vStats used, `/sdk` never called, output byte-identical** |
| vim25 session | `Logout` correctly emitted at the end of the run |

Two completely different vim25 counter names, in two different units, producing the same final metric is the strongest validation I could get without a real vCenter.

### Behaviour before this PR, for comparison

Same mock, original code, across the four response shapes a 9.1 can return:

| `/api/stats/acq-specs` returns | Before | After |
|---|---|---|
| `404` empty body | 💥 `Can't use string ("") as a HASH ref` | ✅ explicit message |
| `404` HTML body | 💥 same Perl crash | ✅ explicit message |
| `200 {}` | ⚠️ `no data for resource host-35 ... at the moment` | ✅ explicit message |
| `404 {"messages":[...]}` | ⚠️ `No stats found` + `creating acq_spec` noise | ✅ explicit message |
| `200 {"acq_specs": []}` (healthy 8.x/9.0) | nominal | **unchanged — no regression** |

### curl examples

Confirming the removal on a real vCenter — `200` on 9.0, `404` on 9.1:

```bash
TOKEN=$(curl -sk -u 'svc-centreon@vsphere.local:<password>' -X POST https://<vcenter>/api/session | tr -d '"')
curl -sk -o /dev/null -w '%{http_code}\n' -H "vmware-api-session-id: $TOKEN" https://<vcenter>/api/stats/counters
curl -sk -o /dev/null -w '%{http_code}\n' -H "vmware-api-session-id: $TOKEN" https://<vcenter>/api/stats/acq-specs
```

Confirming `PerformanceManager` still answers on the same 9.1:

```bash
curl -sk -X POST https://<vcenter>/sdk \
  -H 'Content-Type: text/xml; charset=utf-8' -H 'SOAPAction: "urn:vim25"' \
  -d '<?xml version="1.0" encoding="UTF-8"?>
      <soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns="urn:vim25">
        <soapenv:Body><RetrieveServiceContent>
          <_this type="ServiceInstance">ServiceInstance</_this>
        </RetrieveServiceContent></soapenv:Body>
      </soapenv:Envelope>' | grep -o '<perfManager.*</perfManager>'
```

### Plugin command lines

```bash
# vSphere 8.0 to 9.1, source picked automatically
./centreon_plugins.pl --plugin=apps::vmware::vsphere8::esx::plugin --mode=cpu \
  --hostname=<vcenter> --username=<user> --password=<password> --esx-id=host-35

# force the vim25 source (vSphere 9.1)
./centreon_plugins.pl --plugin=apps::vmware::vsphere8::esx::plugin --mode=cpu \
  --hostname=<vcenter> --username=<user> --password=<password> --esx-id=host-35 \
  --metrics-source=vim25 --verbose
```

`--verbose` shows any alias substitution, which is the first thing to check on a real 9.1.

## Known limitations — please challenge these

1. **The counter name mapping is only verified against mocks.** I do not know which names a real 9.1 `PerformanceManager` advertises. The module degrades safely (exact name → alias → explicit error), but if an alias points at a semantically *close* rather than *equivalent* counter, the value would be wrong without anything flagging it. **The first run against a real 9.1 must compare vim25 values with vStats values from a comparable 9.0.**
2. `intervalId` is pinned to 20 s (real-time) with `maxSample=1`.
3. Only `HostSystem` and `VirtualMachine` entities are handled — which is all the current modes need.
4. **No robot/mockoon integration test yet.** Mockoon was not available in my environment, so the end-to-end verification used the standalone mocks below. Wiring the 9.1 case into `tests/apps/vmware/vsphere8/*/mockoon.json` is the natural follow-up and I am happy to do it in this PR if you prefer it here.

## Unrelated finding, reported for the record

`esx/mode/cpu.pm` converts CPU counters with `* 1000` while `vm/mode/cpu.pm` uses `* 1_000_000` for the same counter family. One of the two is very likely off by a factor of 1000 on the absolute frequency (percentages are unaffected, the ratio cancels out). **I deliberately did not touch it**: `%target_units` reproduces the current behaviour exactly (kHz for hosts, MHz for VMs) so that vStats and vim25 agree. If a mode is corrected, the matching `%target_units` entry must be corrected with it. Worth its own ticket.

## Checklist

- [x] I have followed the **coding style guidelines** provided by Centreon
- [x] I have commented my code, especially **hard-to-understand areas** of the PR
- [x] I have **rebased** my development branch on the base branch (develop)
- [ ] In case of a new plugin, I have created the new packaging directory accordingly — *n/a, existing packages, dependencies updated*
- [x] I have implemented automated tests related to my commits
  - [x] Data used for automated tests are anonymized
- [x] I have reviewed all the help messages in all the .pm files I have modified
  - [x] All sentences begin with a capital letter
  - [x] All sentences end with a period
  - [x] I am able to understand all the help messages
- [x] After having created the PR, I will make sure that all the tests provided in this PR have run and passed

---

<details>
<summary><b>Mock vCenters used for the end-to-end verification</b> (standalone, no dependency — click to expand)</summary>

Save as `fake_vc91.py` (vStats removed, vim25 answers) and `fake_vc90.py` (healthy vStats, records any hit on `/sdk`), then run the plugin against `127.0.0.1:8931`.

`fake_vc91.py` — pass `capacity` or `legacy` to switch the advertised counter catalogue:

```python
import json, re, sys
from http.server import BaseHTTPRequestHandler, HTTPServer

STYLE = sys.argv[1] if len(sys.argv) > 1 else "capacity"

def counters():
    if STYLE == "capacity":
        return [(6, "cpu", "capacity.usage", "average", "kiloHertz"),
                (7, "cpu", "capacity.provisioned", "average", "kiloHertz"),
                (24, "mem", "capacity.usable", "average", "kiloBytes"),
                (25, "mem", "consumed.vms", "average", "kiloBytes"),
                (60, "power", "capacity.usage", "average", "watt")]
    return [(6, "cpu", "usagemhz", "average", "megaHertz"),
            (7, "cpu", "totalCapacity", "average", "megaHertz"),
            (24, "mem", "totalCapacity", "average", "megaBytes"),
            (25, "mem", "consumed", "average", "kiloBytes"),
            (60, "power", "power", "average", "watt")]

SAMPLES = {6: 3200000, 7: 8000000, 24: 134217728, 25: 67108864, 60: 210} if STYLE == "capacity" \
     else {6: 3200, 7: 8000, 24: 131072, 25: 67108864, 60: 210}

ENV = ('<?xml version="1.0" encoding="UTF-8"?>'
       '<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" '
       'xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">'
       '<soapenv:Body>{}</soapenv:Body></soapenv:Envelope>')

class H(BaseHTTPRequestHandler):
    def log_message(self, *a): pass

    def _raw(self, code, body, ctype, extra=None):
        b = body.encode()
        self.send_response(code); self.send_header("Content-Type", ctype)
        if extra: self.send_header(*extra)
        self.send_header("Content-Length", str(len(b))); self.end_headers(); self.wfile.write(b)

    def _json(self, code, payload):
        self._raw(code, json.dumps(payload), "application/json")

    def do_GET(self):
        if self.path.startswith("/api/vcenter/host"):
            return self._json(200, [{"host": "host-35", "name": "esx01.example.tld",
                                     "connection_state": "CONNECTED", "power_state": "POWERED_ON"}])
        if self.path.startswith("/api/stats/"):          # vStats removed in 9.1
            return self._raw(404, "{}", "application/json")
        self._json(404, {"error_type": "NOT_FOUND", "messages": []})

    def do_POST(self):
        if self.path.startswith("/api/session"):
            return self._json(201, "d3adb33f-t0k3n")
        if self.path.startswith("/api/stats/"):
            return self._raw(404, "{}", "application/json")
        if self.path != "/sdk":
            return self._json(404, {"error_type": "NOT_FOUND", "messages": []})

        body = self.rfile.read(int(self.headers.get("Content-Length", 0))).decode()

        if "RetrieveServiceContent" in body:
            return self._raw(200, ENV.format(
                '<RetrieveServiceContentResponse xmlns="urn:vim25"><returnval>'
                '<rootFolder type="Folder">group-d1</rootFolder>'
                '<propertyCollector type="PropertyCollector">propertyCollector</propertyCollector>'
                '<perfManager type="PerformanceManager">PerfMgr</perfManager>'
                '<sessionManager type="SessionManager">SessionManager</sessionManager>'
                '</returnval></RetrieveServiceContentResponse>'), "text/xml",
                ("Set-Cookie", 'vmware_soap_session="52a1b2c3"; Path=/; HttpOnly'))

        if "<Login>" in body:
            return self._raw(200, ENV.format(
                '<LoginResponse xmlns="urn:vim25"><returnval><key>52a1b2c3</key>'
                '<userName>svc</userName></returnval></LoginResponse>'), "text/xml")

        if "RetrievePropertiesEx" in body:
            infos = "".join(
                '<PerfCounterInfo xsi:type="PerfCounterInfo">'
                f'<key>{k}</key>'
                f'<nameInfo><label>l</label><summary>s</summary><key>{name}</key></nameInfo>'
                f'<groupInfo><label>l</label><summary>s</summary><key>{grp}</key></groupInfo>'
                f'<unitInfo><label>l</label><summary>s</summary><key>{unit}</key></unitInfo>'
                f'<rollupType>{roll}</rollupType><statsType>absolute</statsType><level>1</level>'
                '</PerfCounterInfo>' for k, grp, name, roll, unit in counters())
            return self._raw(200, ENV.format(
                '<RetrievePropertiesExResponse xmlns="urn:vim25"><returnval><objects>'
                '<obj type="PerformanceManager">PerfMgr</obj>'
                f'<propSet><name>perfCounter</name><val xsi:type="ArrayOfPerfCounterInfo">{infos}</val></propSet>'
                '</objects></returnval></RetrievePropertiesExResponse>'), "text/xml")

        if "QueryPerf" in body:
            cid = int(re.search(r"<counterId>(\d+)</counterId>", body).group(1))
            val = SAMPLES.get(cid, -1)
            return self._raw(200, ENV.format(
                '<QueryPerfResponse xmlns="urn:vim25"><returnval xsi:type="PerfEntityMetric">'
                '<entity type="HostSystem">host-35</entity>'
                '<sampleInfo><interval>20</interval><timestamp>2026-09-08T10:00:00Z</timestamp></sampleInfo>'
                f'<value xsi:type="PerfMetricIntSeries"><id><counterId>{cid}</counterId><instance></instance></id>'
                f'<value>{val}</value></value>'
                '</returnval></QueryPerfResponse>'), "text/xml")

        if "<Logout>" in body:
            return self._raw(200, ENV.format('<LogoutResponse xmlns="urn:vim25"/>'), "text/xml")

        return self._raw(500, ENV.format(
            '<soapenv:Fault><faultcode>ServerFaultCode</faultcode>'
            '<faultstring>unsupported operation</faultstring></soapenv:Fault>'), "text/xml")

HTTPServer(("127.0.0.1", 8931), H).serve_forever()
```

`fake_vc90.py` — the non-regression mock; it writes `/tmp/sdk_was_called` if the plugin ever touches `/sdk`, which it must not:

```python
import json, time
from http.server import BaseHTTPRequestHandler, HTTPServer

VALUES = {"cpu.capacity.usage.HOST": 3200000, "cpu.capacity.provisioned.HOST": 8000000}

def spec(cid, i):
    return {"counters": {"cid_mid": {"mid": "", "cid": cid}, "set_id": ""},
            "resources": [{"predicate": "EQUAL", "scheme": "moid", "type": "HOST", "id_value": "host-35"}],
            "expiration": int(time.time()) + 864000, "interval": 60, "memo_": "",
            "id": str(200 + i), "status": "ENABLED"}

class H(BaseHTTPRequestHandler):
    def log_message(self, *a): pass

    def _json(self, code, payload):
        b = json.dumps(payload).encode()
        self.send_response(code); self.send_header("Content-Type", "application/json")
        self.send_header("Content-Length", str(len(b))); self.end_headers(); self.wfile.write(b)

    def do_POST(self):
        if self.path.startswith("/api/session"): return self._json(201, "t0k3n")
        if self.path == "/sdk":
            open("/tmp/sdk_was_called", "w").write("yes")
            return self._json(500, {})
        self._json(404, {})

    def do_GET(self):
        if self.path.startswith("/api/vcenter/host"):
            return self._json(200, [{"host": "host-35", "name": "esx01.example.tld"}])
        if self.path.startswith("/api/stats/acq-specs"):
            return self._json(200, {"acq_specs": [spec(c, i) for i, c in enumerate(VALUES)]})
        if self.path.startswith("/api/stats/data/dp"):
            cid = [p[4:] for p in self.path.split("&") if p.startswith("cid=")][0]
            return self._json(200, {"data_points": [{"val": VALUES.get(cid, 0), "ts": int(time.time())}]})
        self._json(404, {})

HTTPServer(("127.0.0.1", 8931), H).serve_forever()
```

Run either mock, then:

```bash
python3 fake_vc91.py capacity &
perl -Isrc src/centreon_plugins.pl --plugin=apps::vmware::vsphere8::esx::plugin --mode=cpu \
  --hostname=127.0.0.1 --port=8931 --proto=http --username=u --password=p \
  --http-backend=lwp --esx-id=host-35 --statefile-dir=/tmp/vsphere8-test
```

</details>

### DIFF
--- a/packaging/centreon-plugin-Virtualization-Vmware8-Esx-Restapi/deb.json
+++ b/packaging/centreon-plugin-Virtualization-Vmware8-Esx-Restapi/deb.json
@@ -1,5 +1,6 @@
 {
     "dependencies": [
-        "libjson-perl"
+        "libjson-perl",
+        "libxml-libxml-perl"
     ]
 }
--- a/packaging/centreon-plugin-Virtualization-Vmware8-Esx-Restapi/rpm.json
+++ b/packaging/centreon-plugin-Virtualization-Vmware8-Esx-Restapi/rpm.json
@@ -1,5 +1,6 @@
 {
     "dependencies": [
-        "perl(JSON)"
+        "perl(JSON)",
+        "perl(XML::LibXML)"
     ]
 }
--- a/packaging/centreon-plugin-Virtualization-Vmware8-Vcenter-Restapi/deb.json
+++ b/packaging/centreon-plugin-Virtualization-Vmware8-Vcenter-Restapi/deb.json
@@ -1,5 +1,6 @@
 {
     "dependencies": [
-        "libjson-perl"
+        "libjson-perl",
+        "libxml-libxml-perl"
     ]
 }
--- a/packaging/centreon-plugin-Virtualization-Vmware8-Vcenter-Restapi/rpm.json
+++ b/packaging/centreon-plugin-Virtualization-Vmware8-Vcenter-Restapi/rpm.json
@@ -1,5 +1,6 @@
 {
     "dependencies": [
-        "perl(JSON)"
+        "perl(JSON)",
+        "perl(XML::LibXML)"
     ]
 }
--- a/packaging/centreon-plugin-Virtualization-Vmware8-Vcsa-Restapi/deb.json
+++ b/packaging/centreon-plugin-Virtualization-Vmware8-Vcsa-Restapi/deb.json
@@ -1,5 +1,6 @@
 {
     "dependencies": [
-        "libjson-perl"
+        "libjson-perl",
+        "libxml-libxml-perl"
     ]
 }
--- a/packaging/centreon-plugin-Virtualization-Vmware8-Vcsa-Restapi/rpm.json
+++ b/packaging/centreon-plugin-Virtualization-Vmware8-Vcsa-Restapi/rpm.json
@@ -1,6 +1,7 @@
 {
     "dependencies": [
         "perl(Date::Parse)",
-        "perl(JSON)"
+        "perl(JSON)",
+        "perl(XML::LibXML)"
     ]
 }
--- a/packaging/centreon-plugin-Virtualization-Vmware8-Vm-Restapi/deb.json
+++ b/packaging/centreon-plugin-Virtualization-Vmware8-Vm-Restapi/deb.json
@@ -1,5 +1,6 @@
 {
     "dependencies": [
-        "libjson-perl"
+        "libjson-perl",
+        "libxml-libxml-perl"
     ]
 }
--- a/packaging/centreon-plugin-Virtualization-Vmware8-Vm-Restapi/rpm.json
+++ b/packaging/centreon-plugin-Virtualization-Vmware8-Vm-Restapi/rpm.json
@@ -1,5 +1,6 @@
 {
     "dependencies": [
-        "perl(JSON)"
+        "perl(JSON)",
+        "perl(XML::LibXML)"
     ]
 }
--- a/src/apps/vmware/vsphere8/custom/api.pm
+++ b/src/apps/vmware/vsphere8/custom/api.pm
@@ -57,6 +57,7 @@ sub new {
                 'password:s'        => { name => 'password' },
                 'vstats-interval:s' => { name => 'vstats_interval',     default => 60 },
                 'vstats-duration:s' => { name => 'vstats_duration',     default => 2764800 }, # 2764800 seconds in 32 days
+                'metrics-source:s'  => { name => 'metrics_source',       default => 'auto' },
                 'timeout:s'         => { name => 'timeout',             default => 10 }
             }
         );
@@ -67,6 +68,9 @@ sub new {
     $self->{http}            = centreon::plugins::http->new(%options, 'default_backend' => 'curl');
     $self->{token_cache}     = centreon::plugins::statefile->new(%options);
     $self->{acq_specs_cache} = centreon::plugins::statefile->new(%options);
+    $self->{vim25_cache}     = centreon::plugins::statefile->new(%options);
+    # kept to build the vim25 fallback lazily, only when a counter is actually needed
+    $self->{custom_options}  = \%options;
 
     return $self;
 }
@@ -90,6 +94,12 @@ sub check_options {
     $self->{password}        = (defined($self->{option_results}->{password})) ? $self->{option_results}->{password} : '';
     $self->{vstats_interval} = $self->{option_results}->{vstats_interval};
     $self->{vstats_duration} = $self->{option_results}->{vstats_duration};
+    $self->{metrics_source}  = $self->{option_results}->{metrics_source} // 'auto';
+
+    if ($self->{metrics_source} !~ /^(auto|vstats|vim25)$/) {
+        $self->{output}->option_exit(short_msg => "Unsupported --metrics-source '" . $self->{metrics_source}
+            . "'. Expected one of: auto, vstats, vim25.");
+    }
 
     if ($self->{hostname} eq '') {
         $self->{output}->option_exit(short_msg => "Need to specify --hostname option.");
@@ -103,6 +113,7 @@ sub check_options {
 
     $self->{token_cache}->check_options(option_results => $self->{option_results});
     $self->{acq_specs_cache}->check_options(option_results => $self->{option_results});
+    $self->{vim25_cache}->check_options(option_results => $self->{option_results});
 
     return 0;
 }
@@ -304,6 +315,60 @@ sub get_vm_guest_identity {
     return $api_response;
 }
 
+# The vStats API (namespace com.vmware.vstats.*) was shipped as Tech Preview in vSphere 8.0
+# and has been REMOVED in vSphere 9.1 ("the vStats APIs (previously in Tech Preview) are
+# removed", VCF 9.1 Product Support Notes). On 9.1 the backend service is no longer
+# registered: /stats/acq-specs and /stats/data/dp fail the same way for GET and POST, while
+# the 9.1 API Explorer still advertises them.
+# Without this guard the response simply has no acq_specs collection, the plugin collects
+# nothing, caches that emptiness and reports "no data at the moment" instead of failing.
+sub vstats_response_is_usable {
+    my ($self, %options) = @_;
+
+    return (ref($options{response}) eq 'HASH' && ref($options{response}->{acq_specs}) eq 'ARRAY') ? 1 : 0;
+}
+
+sub vstats_unavailable_message {
+    my ($self, %options) = @_;
+
+    return "vStats API unusable on '" . $self->{hostname}
+        . "': endpoint '/api/stats/acq-specs' returned no acq_specs collection [http code: '"
+        . $self->{http}->get_code() . "']. The vStats API is Tech Preview in vSphere 8.0 to 9.0 "
+        . "and has been removed in vSphere 9.1, therefore performance counters cannot be "
+        . "collected through it. Use --metrics-source=vim25 to read them from the "
+        . "PerformanceManager SOAP API instead. Modes that do not rely on performance "
+        . "counters (status, health, count, list...) are not affected.";
+}
+
+# vim25 PerformanceManager fallback, built only when a counter is actually requested
+sub vim25 {
+    my ($self, %options) = @_;
+
+    return $self->{vim25} if (defined($self->{vim25}));
+
+    centreon::plugins::misc::mymodule_load(
+        output    => $self->{output},
+        module    => 'apps::vmware::vsphere8::custom::vim25',
+        error_msg => "Cannot load module 'apps::vmware::vsphere8::custom::vim25'."
+    );
+
+    $self->{vim25} = apps::vmware::vsphere8::custom::vim25->new(
+        %{$self->{custom_options}},
+        noptions => 1,
+        output   => $self->{output},
+        cache    => $self->{vim25_cache},
+        hostname => $self->{hostname},
+        port     => $self->{port},
+        proto    => $self->{proto},
+        username => $self->{username},
+        password => $self->{password},
+        timeout  => $self->{timeout}
+    );
+    $self->{vim25}->set_options(option_results => $self->{option_results});
+
+    return $self->{vim25};
+}
+
 sub get_all_acq_specs {
     my ($self, %options) = @_;
 
@@ -319,21 +384,32 @@ sub get_all_acq_specs {
     }
     # Get all acq specs (first page)
     my $response =  $self->request_api(endpoint => '/stats/acq-specs') ;
+    # give up instead of collecting nothing when the vStats API is not usable;
+    # get_stats() decides whether to fall back to vim25 or to fail explicitly
+    if (!$self->vstats_response_is_usable(response => $response)) {
+        $self->{vstats_unavailable} = 1;
+        return undef;
+    }
 
     # store only acq_specs related to the considered resource
-    push @{$self->{all_acq_specs}}, grep {$_->{resources}->[0]->{id_value} eq $options{rsrc_id}} @{$response->{acq_specs}};
+    my @acq_specs = grep {$_->{resources}->[0]->{id_value} eq $options{rsrc_id}} @{$response->{acq_specs}};
     # If the whole acq-specs takes more than one page, the API will return a "next" value
     while ($response->{next}) {
         $response = $self->request_api(endpoint => '/stats/acq-specs', get_param => [ 'page=' . $response->{next} ] );
+        if (!$self->vstats_response_is_usable(response => $response)) {
+            $self->{vstats_unavailable} = 1;
+            return undef;
+        }
         # store only acq_specs related to the considered resource
-        push @{$self->{all_acq_specs}}, grep {$_->{resources}->[0]->{id_value} eq $options{rsrc_id}} @{$response->{acq_specs}};
+        push @acq_specs, grep {$_->{resources}->[0]->{id_value} eq $options{rsrc_id}} @{$response->{acq_specs}};
     }
 
+    # only a well-formed list is kept and cached: this prevents a corrupted or empty
+    # statefile from making every subsequent run collect nothing
+    $self->{all_acq_specs} = \@acq_specs;
     # store it in the cache for future runs
     $self->{acq_specs_cache}->write(data => { updated => time(), acq_specs => $self->{all_acq_specs} });
-    # in some cases the statefile was corrupted and the plugin was stuck
-    # only return a result if it looks ok, else we store an empty array
-    $self->{all_acq_specs} = [] unless ref($self->{all_acq_specs}) eq 'ARRAY' && @{$self->{all_acq_specs}};
+
     return $self->{all_acq_specs};
 }
 
@@ -427,6 +503,7 @@ sub get_acq_spec {
 
     # If it is not available in cache call get_all_acq_specs()
     my $acq_specs = $self->get_all_acq_specs(%options);
+    return undef if (!defined($acq_specs));
     for my $spec (@$acq_specs) {
         # Ignore acq_specs not related to the counter_id
         next if ($options{cid} ne $spec->{counters}->{cid_mid}->{cid});
@@ -446,6 +523,9 @@ sub check_acq_spec {
     my ($self, %options) = @_;
 
     my $acq_spec = $self->get_acq_spec(%options);
+
+    # do not try to create an acq_spec on a vCenter where vStats no longer exists
+    return undef if ($self->{vstats_unavailable});
 
     if ( !defined($acq_spec) ) {
         # acq_spec not found => we need to create it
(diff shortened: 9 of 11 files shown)


julienmathis:
You are right, and the miss was worse than "not automatic": the fix simply did not fire at all for those pollers. Fixed in 2408586.

## Confirming the diagnosis

`get_all_acq_specs()` returned the cached list at line 379-384, **before** reaching the detection added at line 389. So a poller that had already run against 9.1 short-circuited the whole fix.

Reproduced with the 9.1 mock — pre-fix code first, then the fixed code over the untouched statefile:

```
### 1. 9.1, code from develop -> poisons its own cache
UNKNOWN: no data for resource host-35 counter cpu.capacity.provisioned.HOST at the moment.
cache: {"acq_specs":[],"updated":1788946489}

### 2. same poller, fix applied, poisoned cache left in place
UNKNOWN: no data for resource host-35 counter cpu.capacity.provisioned.HOST at the moment.
The counter cpu.capacity.provisioned.HOST was not recorded for resource host-35 before. It will now (creating acq_spec).
```

Still silent, and still trying to create acquisition specifications against an API that no longer exists. Exactly what you described.

## Why not delete the statefiles by age

Deleting `/var/lib/centreon/centplugins/vsphere8_api_acq_specs_*` past a certain age was considered and rejected, for three reasons:

1. **Age does not identify the defect.** A legitimately populated statefile gets old too. Sweeping by age would discard healthy caches, re-query the API and re-create acquisition specifications on vCenters that had nothing wrong with them.
2. **It leaves a window of silence.** Whatever threshold is picked, an affected poller keeps reporting nothing until it elapses. The population we are trying to help is precisely the one that should not have to wait.
3. **A plugin has no business globbing and unlinking in that directory.** It is shared by every plugin's statefiles, and a glob-and-delete running as the poller user is a poor trade for a problem that has a precise fix.

## What was done instead

An empty or malformed cached list is now **a cache miss, never an answer**. The API is queried again, the removal is detected, and the vim25 source takes over. No threshold, no sweeping, no manual action: an affected poller heals on its very next check, and the stale statefile is simply overwritten by the next successful collection.

Same reproduction, with the fix, statefile untouched:

```
### 2. same poller, fix applied, poisoned cache left in place
OK: CPU average usage is 40.00 %, used frequency is 3200000 kHz | ...

### and with --metrics-source=vstats (no fallback)
UNKNOWN: vStats API unusable on '<vcenter>': endpoint '/api/stats/acq-specs' returned no
acq_specs collection [http code: '404']. The vStats API is Tech Preview in vSphere 8.0 to
9.0 and has been removed in vSphere 9.1 ...
```

## Your comment surfaced a second, pre-existing defect

Chasing this down showed the statefile was **written mid-run with a partial view**, and creating a specification did not invalidate it. So on a *healthy* vCenter, every later run re-created the specifications missing from that snapshot, accumulating them on the customer's vCenter.

Traced run by run against a mock that persists what it creates, on a two-counter mode:

```
### develop
run 1 : POST=2  cache=1 spec(s): ['cpu.capacity.provisioned.HOST']
run 2 : POST=3  cache=1 spec(s): ['cpu.capacity.provisioned.HOST']
run 3 : POST=4  cache=1 spec(s): ['cpu.capacity.provisioned.HOST']
```

One redundant `POST /stats/acq-specs` per run, forever, because the cached snapshot never contained the second counter.

`create_acq_spec()` and `extend_acq_spec()` now share `invalidate_acq_specs()`, and a list read while specifications are being created is no longer persisted:

```
### with the fix
run 1 : POST=2  cache=absent
run 2 : POST=2  cache=2
run 3 : POST=2  cache=2
run 4 : POST=2  cache=2
```

Two creations total, and the cache converges to the complete list from the second run on. That part is not strictly what you asked about — say the word and I split it into its own commit or its own PR if you would rather keep this one narrow.

## Tests

**53 assertions, up from 38.** The 15 added cover this exact scenario:

- an empty cached list does not short-circuit the detection, and the removal is still detected
- nothing is written back when vStats is unusable
- a malformed cached value (`undef`, a string, a hash) is treated as a cache miss
- a populated cache is still trusted, and triggers no request
- a non-empty API answer is cached, an empty one is not
- an empty but well-formed API answer is not an error (healthy vCenter with no specification yet)
- `invalidate_acq_specs()` drops the stored list and flags it stale

Non-regression re-verified: a fresh 9.1 poller falls back to vim25, an already-affected 9.1 poller heals, and a healthy 9.0 still uses vStats without ever touching `/sdk`.

